PR #16, opened 2024-01-23
Add cartesian product decorator for benchmarks

### LINKED ISSUE #5: Add cartesian product decorator
`@nnbench.parametrize` calculates a multi-way "dot product" of all arguments. However, a cartesian product (i.e. a benchmark grid over the parameters) is also useful.

Demo:

```python

@nnbench.product(
    a=[1, 2, 3, 4, 5],
    b=[True, False],
)
def double_if_b(a: int, b: bool) -> int:
    return a * 2 if b else a
```

This decorator should produce a list of `len(a) * len(b) = 10` benchmarks, and obviously saves space over a 10-way parametrization.

nicholasjng:
Adds the third core decorator for parametrizing benchmarks, which is the cartesian product of variable collections. It is exposed as `nnbench.product`.

Compared to the "inner product" approach of `nnbench.parametrize`, which consumes an iterable of dictionaries, this decorator consumes a keyworded collection of iterables, which it then binds to the benchmark function. This results in an "outer product" with a number of benchmarks that is equal to the products of the numbers of values in each iterable.

However, since `functools.partial` binds parameters lazily, users do not immediately get feedback on whether they supplied the correct argument names in the iterable variable keyword arguments. This needs to be improved, by either implementing a check inside the decorator or by typing it to the decorated benchmark function's param spec (Python 3.11+).

Closes #5.

### DIFF
--- a/src/nnbench/__init__.py
+++ b/src/nnbench/__init__.py
@@ -9,7 +9,7 @@ except PackageNotFoundError:
     pass
 
 # TODO: This naming is unfortunate
-from .core import benchmark, parametrize
+from .core import benchmark, parametrize, product
 from .reporter import BaseReporter, register_reporter
 from .types import Benchmark, Params
 
--- a/src/nnbench/core.py
+++ b/src/nnbench/core.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import itertools
 from functools import partial, update_wrapper
 from typing import Any, Callable, Iterable, overload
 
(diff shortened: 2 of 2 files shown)


nicholasjng:
Usage:

```py
import nnbench

@nnbench.product(a=[1, 2, 3], b=[4, 5, 6])
def prod(a: int, b: int) -> int:
    return a * b
```

This results in a list of `len(a) * len(b) = 9` benchmarks, one for each possible combination of a and b.

PR #18, opened 2024-01-24
Handle untyped interfaces in `_check`

maxmynter:
Guard match checks of types in function signature and supplied via the `params` kwarg in the `benchmark` decorators `_check` method against untyped function signatures. 

If untyped, we do not check for matching types. We also log a warning to the `stdout` of the `logger` indicating the untyped parameter and function name.

### DIFF
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -23,6 +23,10 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
     for bm in benchmarks:
         for name, param in inspect.signature(bm.fn).parameters.items():
             param_type = param.annotation
+
+            if param.annotation == inspect.Parameter.empty:
+                logger.warn(f"Warning: Parameter {name!r} in benchmark {bm.fn.__name__} is untyped.")
+
             if name in benchmark_interface and benchmark_interface[name].annotation != param_type:
                 orig_type = benchmark_interface[name].annotation
                 raise TypeError(
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -29,7 +29,7 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
 
             param_type = param.annotation
 
-            if param.annotation == inspect.Parameter.empty:
+            if param_type == inspect.Parameter.empty:
                 logger.debug(f"Found untyped parameter {name!r} in benchmark {bm.fn.__name__!r}.")
 
             if name in benchmark_interface and benchmark_interface[name].annotation != param_type:
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -29,8 +29,8 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
                     f"Warning: Parameter {name!r} in benchmark '{bm.fn.__name__}' is untyped."
                 )
 
-            if name in benchmark_interface and benchmark_interface[name].annotation != param_type:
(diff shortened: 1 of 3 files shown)


src/nnbench/runner.py
```diff
@@ -23,6 +23,12 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
     for bm in benchmarks:
         for name, param in inspect.signature(bm.fn).parameters.items():
             param_type = param.annotation
+
+            if param.annotation == inspect.Parameter.empty:
+                logger.warn(
+                    f"Warning: Parameter {name!r} in benchmark '{bm.fn.__name__}' is untyped."
+                )
```
nicholasjng, line 30: I think this should be a debug log - it's more of a convenience for the user if they want to understand why something is failing, and we will always and unconditionally advertise typing benchmarks completely.

```suggestion
            if param_type == inspect.Parameter.empty:
                logger.debug(
                    f"Found untyped parameter {name!r} in benchmark {bm.fn.__name__!r}."
                )
```

src/nnbench/runner.py
```diff
@@ -32,7 +38,9 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
     for name, param in benchmark_interface.items():
         if name not in param_types and param.default == inspect.Parameter.empty:
             raise ValueError(f"missing value for required parameter {name!r}")
-        if not issubclass(param_types[name], param.annotation):
+        if not param.annotation == inspect.Parameter.empty and not issubclass(
+            param_types[name], param.annotation
+        ):  # only check type match if type supplied
```
nicholasjng, line 43:
```suggestion
        # skip the check if benchmark interface is untyped
        if param.annotation == inspect.Parameter.empty:
            continue
        if not issubclass(param_types[name], param.annotation):
```

nicholasjng (commented):
With this change, the following should be a failure (non-unique types):

```py
import nnbench

@nnbench.benchmark
def increment(value):
    return value + 1


@nnbench.benchmark
def double(value: int) -> int:
    return value * 2
```

Is this expected? Are we fine with this?

(We should check that the resulting type in the error message is still sensible, though. Is it `<class 'empty'>` by any chance?)

maxmynter:
> (We should check that the resulting type in the error message is still sensible, though. Is it `<class 'empty'>` by any chance?)

It is `<class 'inspect._empty'>.

nicholasjng:
> > (We should check that the resulting type in the error message is still sensible, though. Is it `<class 'empty'>` by any chance?)
> 
> It is `<class 'inspect._empty'>`.

It's fine for now. We can think about making it show up as `<untyped>` or something, but that requires some more effort, and would only be relevant in the case of an error, i.e. just before raising the exception.

maxmynter:
i would argue that your example above should throw an error. Untyped and typed might very well be non-unique types. We cannot handle different types for the same variable name and if untyped cannot ensure it is the right one. So it's either on the user to keep everything untyped and ensure the right value is passed. Or the absence of the type is an oversight in which case the error is a welcome reminder.

src/nnbench/runner.py
```diff
@@ -28,15 +28,25 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
                 )
 
             param_type = param.annotation
+
+            if param.annotation == inspect.Parameter.empty:
```
nicholasjng, line 32:
```suggestion
            if param_type == inspect.Parameter.empty:
```

nicholasjng (approved):

PR #19, opened 2024-01-24
Add warning for supplied and default parameter collision

maxmynter:
tbd

### DIFF
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -22,6 +22,11 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
     benchmark_interface: dict[str, inspect.Parameter] = {}
     for bm in benchmarks:
         for name, param in inspect.signature(bm.fn).parameters.items():
+            if not param.default == inspect.Parameter.empty and param in params:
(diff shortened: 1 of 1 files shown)


src/nnbench/runner.py
```diff
@@ -22,6 +22,11 @@ def _check(params: dict[str, Any], benchmarks: list[Benchmark]) -> None:
     benchmark_interface: dict[str, inspect.Parameter] = {}
     for bm in benchmarks:
         for name, param in inspect.signature(bm.fn).parameters.items():
+            if not param.default == inspect.Parameter.empty and name in params:
+                logger.warning(
+                    f"received parameter {name!r}={params[name]}, overwrite the default ({name!r}={param.default}) in function '{bm.fn.__name__}'"
+                )
```
nicholasjng, line 28:
```suggestion
            if name in params and param.default != inspect.Parameter.empty:
                logger.debug(
                    f"using value {params[name]} instead of default {param.default} for parameter {name!r} in function {bm.fn.__name__!r}"
                )
```

Might have to fix the tests accordingly.

nicholasjng (commented):
Nice work!

PR #20, opened 2024-01-24
Add interface class member to Benchmark class

maxmynter:
Add an `Interface` immutable dataclass to the `Benchmark` class that contains the signature information of a function. I.e.: 

- Keys
- Values
- Key - Value Pairs 
- Defaults

### DIFF
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -108,6 +108,23 @@ class Benchmark:
 
 @dataclass(frozen=True)
 class Interface:
+    """
+    Data model representing a function's interface. An instance of this class
+    is created using the `from_callable` class method.
+
+    Parameters:
+    ----------
+    names : tuple[str, ...]
+        Names of the function parameters.
+    types : tuple[type, ...]
+        Types of the function parameters.
+    variables : tuple[Variable, ...]
+        A tuple of tuples, where each inner tuple contains the parameter name and type.
+    defaults : dict[str, Any]
+        A dictionary mapping the parameters names to default values.
+        Only contains parameters with default values.
+    """
+
     names: tuple[str, ...]
     types: tuple[type, ...]
     variables: tuple[Variable, ...]
@@ -115,6 +132,9 @@ class Interface:
 
     @classmethod
     def from_callable(cls, fn: Callable) -> Interface:
+        """
+        Creates an Interface instance from a given callable.
+        """
         sig = inspect.signature(fn)
         return cls(
             tuple(sig.parameters.keys()),
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -1,9 +1,10 @@
 """Useful type interfaces to override/subclass in benchmarking workflows."""
 from __future__ import annotations
 
+import inspect
 import os
 from dataclasses import dataclass, field
-from typing import Any, Callable, Generic, TypedDict, TypeVar
+from typing import Any, Callable, Dict, Generic, Tuple, TypedDict, TypeVar
 
 T = TypeVar("T")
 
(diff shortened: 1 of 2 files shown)


src/nnbench/types.py
```diff
@@ -1,9 +1,10 @@
 """Useful type interfaces to override/subclass in benchmarking workflows."""
 from __future__ import annotations
 
+import inspect
 import os
 from dataclasses import dataclass, field
-from typing import Any, Callable, Generic, TypedDict, TypeVar
+from typing import Any, Callable, Dict, Generic, Tuple, TypedDict, TypeVar
```
nicholasjng, line 7: Please use builtins (`tuple, dict`) everywhere.

src/nnbench/types.py
```diff
@@ -93,9 +94,24 @@ class Benchmark:
...
+        vartypes: Tuple[inspect.Parameter, ...] | None = field(default=None)
+        varitems: Tuple[Tuple[str, inspect.Parameter], ...] | None = field(default=None)
+        defaults: Dict[str, Any] | None = field(default=None)
+
+        def __post_init__(self) -> None:
+            sig = inspect.signature(self.fn)
+            super().__setattr__("varnames", tuple(sig.parameters.keys()))
+            super().__setattr__("vartypes", tuple(sig.parameters.values()))
+            super().__setattr__("varitems", tuple(sig.parameters.items()))
+            super().__setattr__("defaults", {n: p.default for n, p in sig.parameters.items()})
```
nicholasjng, line 110: This together with `fn` appearing in the initializer is a bit unfortunate. How about we copy `inspect.Signature` in that regard by using a classmethod?

```py
@dataclass(frozen=True)
class Interface:
    # no fn slot here.
    varnames: tuple[str, ...]  # <- these should never be None.
    vartypes: tuple[type, ...] # <- ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...

    @classmethod
    def from_callable(cls, fn: Callable) -> "Interface":
        ... # fill all of the above slots.
        return cls(...)
```

Notice the `tuple[type, ...]` on `vartypes` - the point is to grab the metadata from the `Parameter`s, not save the parameters themselves (otherwise we would just keep rolling with the `inspect.Signature`!)

I would even go so far as to not make it default-constructible (i.e. leave out `__init__`), since this is internal-only and we're just refining some of the info from the `inspect.Signature`.

nicholasjng (commented):
That's a great test case!

Let's move `Interface` to the `nnbench.types` top-level though - we don't (presumably) need to use it anywhere except the `Benchmark` dataclass for now.

src/nnbench/types.py
nicholasjng: I just realized I screwed up the `vartypes` typing in the issue, so that's on me.

src/nnbench/types.py
```diff
@@ -98,4 +99,22 @@ def __post_init__(self):
...
+    varitems: tuple[tuple[str, inspect.Parameter], ...]
+    defaults: dict[str, Any]
+
+    @classmethod
+    def from_callable(cls, fn: Callable) -> Interface:
+        sig = inspect.signature(fn)
+        return cls(
+            tuple(sig.parameters.keys()),
+            tuple(p.annotation for p in sig.parameters.values()),
+            tuple(sig.parameters.items()),
```
nicholasjng, line 118:
```suggestion
            tuple((k, v.annotation) for k, v in sig.parameters.items()),
```

tests/test_benchmark_cls.py
```diff
@@ -0,0 +1,32 @@
...
+    assert interface.varnames == ("a", "b", "c", "d")
+    assert interface.vartypes == (
+        int,
+        inspect._empty,
+        str,
+        float,
+    )
+
+    varitems = [(param.name, param.annotation) for name, param in interface.varitems]
+    assert varitems == [("a", int), ("b", inspect._empty), ("c", str), ("d", float)]
```
nicholasjng, line 31: With the above suggestion, you can just directly `assert interface.varitems()`.

Also, I would prefer a naming like `interface.{names, types, variables}`.

src/nnbench/types.py
```diff
@@ -98,4 +99,22 @@ def __post_init__(self):
...
             super().__setattr__("name", name)
-        # TODO: Parse interface using `inspect`, attach to the class
+            super().__setattr__("interface", Interface.from_callable(self.fn))
+
+
+@dataclass(frozen=True)
+class Interface:
+    varnames: tuple[str, ...]
+    vartypes: tuple[type, ...]
+    varitems: tuple[tuple[str, inspect.Parameter], ...]
```
nicholasjng, line 109:
```suggestion
    varitems: tuple[tuple[str, type], ...]
```

(Maybe you can even do a type alias `Variable = tuple[str, type]` and then go `variables: tuple[Variable, ...]` with the suggestion below.)

nicholasjng:
Sorry I forgot, you also need to add `interface` to the `Benchmark` class as a field like so (otherwise it will not show up in the typing):

```py
class Benchmark:
    """
    ...
    interface: Interface
        The benchmark function's interface, set in the post-init hook.
    """
    ...
    interface: Interface = field(init=False, repr=False)
```

nicholasjng (approved):
Nice job! You can merge directly after attaching the interface to the `Benchmark`.

Issue #13, opened 2024-01-23
Improve test directory organization


nicholasjng:
We have two different test directories with benchmarks in them, because we are testing two different aspects of benchmark collection.

The number of tests for collection will probably increase, so we should come up with a way to improve the test directory setup. An obvious way would be to consolidate all benchmark files into one directory, tag these benchmarks with different test workloads, and then filter by these tags in their respective unit tests.

This has the obvious advantage that we get used to the concept of filtering benchmarks by tags, and have a more intuitive understanding of benchmark collection by our package.

PR #23, opened 2024-01-25
Reorganize test directory structure

maxmynter:
- Move all functions for benchmarks into a single file in the `testproject` directory.
- Except subdirectory with single file for runner discovery test such that we can check the amount of discovered files and still add new functions or files in the `testproject` directory
- Add tags to functions and filter on them in function collection of tests. As such, we have direct association and control over the available functions associated with a test while simultaneously keeping the flexibility to add more functions. Also this serves as a smoke test for the tags.  

- Small bugfix: Pass tags in recursive call of `collect` of benchmark.

Closes https://github.com/aai-institute/nnbench/issues/13


nicholasjng:
Nice work, but I think we should advertise semantic structure of the benchmarks on the file level as well.

By that, I mean that similar benchmarks should be grouped in the same file, with supporting tags like you did in this PR. For instance, take a look at how JAX does it: https://github.com/google/jax/tree/main/benchmarks

By looking at that tree, you would know immediately which file to check if you were tasked with e.g. looking up a differential equation solver benchmark or a matrix inversion benchmark.

My proposal: Rename the `testproject` folder to `benchmarks` (or testbenchmarks or something), keep the semantic tags, and make a file for each larger test (e.g. one for pure `@nnbench.benchmark` collection (+failure cases?), one for `@nnbench.parametrize` collection, etc.)

Then, in each respective test, collect on the single corresponding benchmark file only.

nicholasjng (approved):

PR #25, opened 2024-01-26
Add usage example to README.md

maxmynter:
Added a usage example to `README.md`.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -3,4 +3,64 @@
 Welcome! This quickstart guide will convey the basics needed to use nnbench.
 You will define a benchmark, initialize a runner and reporter, and execute the benchmark, obtaining the results in the console in tabular format.
 
-TODO: Add content.
+## A short scikit-learn model benchmark 
+
+In the following simple example, we put the training and benchmarking logic in the same file. For more complex workloads, we recommend structuring your code into multiple files to improve project organization, similarly to unit tests.
+See the user guides (TODO: Add guides) at the bottom of this page for inspiration.
+
+```python
+from sklearn.datasets import load_iris
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.model_selection import train_test_split
+
+data = load_iris()
+X, y = data.data, data.target
+
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.3)
+
+model = RandomForestClassifier()
+model.fit(X_train, y_train)
+```
+
+To benchmark your model, you encapsulate the benchmark code into a function and apply the `@benchmark` decorator. 
+This marks the function for collection to our benchmark runner later.
+
+```python
+import nnbench
+import numpy as np
+from sklearn import base, metrics
+
+
+@nnbench.benchmark()
+def accuracy(model: base.BaseEstimator, y_test: np.ndarray, y_pred: np.ndarray) -> float:
+    accuracy = metrics.accuracy_score(y_test, y_pred)
+    return accuracy
+```
+
+Now we can instantiate a benchmark runner to collect and run the accuracy benchmark.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,77 @@ poetry add git+https://github.com/aai-institute/nnbench.git
...
+data = load_iris()
+X, y = data.data, data.target
+
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.3)
+
+model = RandomForestClassifier()
+model.fit(X_train, y_train)
+```
+
+To benchmark your model, you encapsulate the benchmark code into a function and assign the `@benchmark` decorator. This let's our benchmark class collect and later execute the function.
```
nicholasjng, line 46:
```suggestion
To benchmark your model, you encapsulate the benchmark code into a function and apply the `@benchmark` decorator. This marks the function for collection to our benchmark runner later.
```

README.md
```diff
@@ -24,6 +24,77 @@ poetry add git+https://github.com/aai-institute/nnbench.git
...
+from training import model, X_test, y_test
+
+
+PATH_TO_BENCHMARK_DIR = pathlib.Path('.').resolve()
+
+
+class ExampleRunner(runner.AbstractBenchmarkRunner):
+    # Here you could integrate your custom benchmark handling logic.
+    # We keep it empty in this example
+    pass
```
nicholasjng, line 76: You did the right thing here, but we shouldn't advertise this in the readme. Therefore, we should probably rename the abstract class to just `BenchmarkRunner`.

README.md
```diff
@@ -24,6 +24,77 @@ poetry add git+https://github.com/aai-institute/nnbench.git
...
+
+
+r = ExampleRunner()
+r.collect()
+y_pred = model.predict(X_test)
+
+result = r.run(PATH_TO_BENCHMARK_DIR, tags=("accuracy",),  params={"model": model, "y_pred": y_pred, "y_test": y_test})
```
nicholasjng, line 83: This should probably work with "benchmark.py"? Then you can save the `PATH_TO_...` definition, and drop the tags. (Less is more!)

README.md
```diff
@@ -24,6 +24,77 @@ poetry add git+https://github.com/aai-institute/nnbench.git
...
+
+r.report(to='console', result=result)
+```
+The output then looks like this
```
nicholasjng, line 87:
```suggestion
The resulting output might look like this:
```

nicholasjng (commented):
Great job!

README.md
maxmynter: it does with `"./benchmark.py"`.

README.md
```diff
@@ -24,6 +24,69 @@ poetry add git+https://github.com/aai-institute/nnbench.git
 
 A PyPI package release is planned for the future.
 
+## Usage Example
+
+An example repository to benchmark a ML model contains three files. The `training.py` contains the code to train the model, the `benchmark.py` contains the benchmark functions, and the `run_benchmark.py` the benchmark execution logic.
```
nicholasjng, line 29: For this example it's fine, but I'm not sure if we want to advertise that file structure regardless.

It makes sense if we think about benchmarks in a MLOps pipeline way. Question is, do we still include it in the introductory example to only showcase the basic application?

I think this example should serve as the Quickstart document, and be a single-file monolith to convey the simplicity of the package.

Then, if the user is interested, we link to user guides (to be written) on file structure/organization, tagging, subclassing runners,reporters,etcpp. What do you think?

nicholasjng (approved):
Nice! Check this in and then let's talk about restructuring later. 

Please rebase as two commits , one for the example and one for the runner rename.

To do that, you can either use `git rebase -i` or `git reset main` and then craft the two commits manually by adding the respective file set (changes should be orthogonal in this case).

nicholasjng (approved):

PR #39, opened 2024-01-31
Add benchmark organization guide

### LINKED ISSUE #32: Add user guide: Best practices on modularizing benchmark code
Add user guide to the documentation:
- [ ] Best practices on modularizing benchmark code (split into different files, benchmark organization based on semantic groups, reusing setup/teardown code, ...)

nicholasjng:
Contains the two most important structuring tips - refactoring into a proper directory and semantic grouping - in a user guide document.

Also introduces tags as another layer of organization that allows grouping across different module files.

I based this on the MNIST example #11 to take over those changes for the docs organization, so that needs to go in before this.

Closes #32.

### DIFF
--- a/docs/guides/organization.md
+++ b/docs/guides/organization.md
@@ -1,0 +1,83 @@
+# How to efficiently organize benchmark code
+
+To efficiently organize benchmarks and keeping your setup modular, you can follow a few guidelines.
+
+## Tip 1: Separate benchmarks from project code
+
+This tip is well known from other software development practices such as unit testing.
+To improve project organization, consider splitting off your benchmarks into their own modules or even directories, if you have multiple benchmark workloads.
+
+An example project layout can look like this, with benchmarks as a separate directory at the top-level:
+
+```
+my-project/
+├── benchmarks/ # <- contains all benchmarking Python files.
+├── docs/
+├── src/
+├── .pre-commit-config.yaml
+├── pyproject.toml
+├── README.md
+└── ...
+```
+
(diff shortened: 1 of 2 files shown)


docs/guides/organization.md
```diff
@@ -0,0 +1,83 @@
+# How to efficiently organize benchmark code
+
+To efficiently organize benchmarks and keeping your setup modular, you can follow a few guidelines.
```
maxmynter, line 3:
```suggestion
To efficiently organize benchmarks and keeping your setup modular, we propose the following guidelines.
```

docs/guides/organization.md
```diff
@@ -0,0 +1,83 @@
...
+## Tip 3: Attach tags to benchmarks for selective filtering
+
+For structuring benchmarks within files, you can also use **tags**, which are tuples of strings attached to a benchmark:
+
+```python
+# benchmarks/data_quality.py
+import nnbench
+
+
+@nnbench.benchmark(tags=("foo",))
```
maxmynter, line 58: Should we add a small note like:

> Note the trailing comma. This is necessary for Python to correctly parse single entry tuples as such. Without it, the value will be interpreted as a string, causing a type error.

maxmynter (commented):
Nice. Short, to the point but valuable.

docs/guides/organization.md
nicholasjng: That's not really our issue, IDEs complain when you leave out the comma as far as I know.

PR #43, opened 2024-02-01
Make first argument to `@nnbench.parametrize` a single iterable

nicholasjng:
The stararg change made the decorator unnecessarily inflexible, as no more plain iterables were allowed.

Adjusts the parametrization example to take a raw tuple instead of unraveled arguments.


nicholasjng:
Forgot to mention in the commit message, this also removes the leftover return type hints of the `parametrize` and `product` decorators that were stale when we removed the `func` arguments.

PR #51, opened 2024-02-01
Add namegen argument to `@nnbench.{parametrize,product}`

### LINKED ISSUE #40: Add `namegen` argument to `@nnbench.{parametrize,product}`
Similarly to defining a custom name for a single benchmark, we should give an option to give a custom name to each benchmark of a family defined by the `@nnbench.parametrize` and `@nnbench.product` decorators.

To prevent collisions, this should not be a static string, but a callable taking the input parameters as a dictionary and returning a string that will subsequently be set as the name of the benchmark for the respective parameter instance.

nicholasjng:
Switches out the generated name for a custom name, with the default staying the same as before. The default (function name plus underscore- joined list of keyword arguments) was refactored into a module-private name generator.

Closes #40.

### DIFF
--- a/src/nnbench/core.py
+++ b/src/nnbench/core.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import itertools
+import warnings
 from functools import partial, update_wrapper
 from typing import Any, Callable, Iterable, overload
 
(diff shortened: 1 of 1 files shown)


nicholasjng:
NB: This does not require the generated names to be unique, even though for all intents and purposes they should be. If we want to add this, options are to either warn or error on a duplicate (requires keeping track of all names in the created decorator).

nicholasjng:
I'm not patient enough to refactor this right now, so I will revisit it later.

Issue #49, opened 2024-02-01
Finalize names of user-facing abstract classes


nicholasjng:
Names should be clear and unique, communicating what they do and where/how to use them.

Candidates for refactoring:

* nnbench.Params -> nnbench.Parameters.
* nnbench.Benchmark is a suitable name, but should drop out of the top-level module namespace.
* nnbench.BaseReporter -> nnbench.BenchmarkReporter
* nnbench.types.BenchmarkResult -> nnbench.types.BenchmarkRecord? nnbench.types.Record?

PR #53, opened 2024-02-01
Finalize names

maxmynter:
Closes https://github.com/aai-institute/nnbench/issues/49

- Decided on namechange `BenchmarkResult` -> `Record` since the latter is shorter and we do not record anything besides benchmarks.

- Namechange of `BenchmarkReporter` adressed in https://github.com/aai-institute/nnbench/pull/52


nicholasjng (commented):
I'd like it to be `BenchmarkRecord`, everything else looks perfect.

EDIT: Except for the fact that the arguments are also still named `result`, so if you could change that too, that would be nice.

Issue #50, opened 2024-02-01
Add test coverage for nnbench core abstractions


nicholasjng:
We decided to forego test coverage in the past, but we should add it back to establish confidence into our first shipped release.

Concrete steps:

- [x] Core decorator tests: Check names, interfaces, fn attributes (docstrings, defaults etc.)
- [x] Collection tests: Number of benchmarks, tag matching, multifile/directory collection, typecheck fails.
- [x] Context: Collection, failure cases (non-unique keys)

PR #54, opened 2024-02-01
Add test coverage

maxmynter:
Closes https://github.com/aai-institute/nnbench/issues/50

Tests for collection were already implemeted for the most part (`test_runner.py` and `test_argcheck.py`)

## Open Problems
- For me locally, the assertions like `assert product_benchmark[0].fn.keywords == {"iter1": 1, "iter2": 'a'}` in the core decorator tests come with MyPy errors 
```bash
"Callable[..., Any]" has no attribute "keywords"Mypy[attr-defined](https://mypy.readthedocs.io/en/latest/_refs.html#code-attr-defined)
```


tests/test_decorators.py
```diff
@@ -0,0 +1,65 @@
...
+def test_parametrize():
+    @parametrize([{"param": 1}, {"param": 2}])
+    def parametrized_benchmark(param: str) -> str:
+        return param
+
+    assert len(parametrized_benchmark) == 2
+    assert parametrized_benchmark[0].fn.keywords["param"] == 1  # type: ignore
+    assert parametrized_benchmark[0].fn() == 1
+    assert parametrized_benchmark[1].fn.keywords["param"] == 2  # type: ignore
+    assert parametrized_benchmark[1].fn() == 2
```
nicholasjng, line 33: RIP, those are the wrong arg types. Thank you for catching this!

tests/test_decorators.py
```diff
@@ -0,0 +1,65 @@
...
+        return iter1, iter2
+
+    assert len(product_benchmark) == 4
+    assert product_benchmark[0].fn.keywords == {"iter1": 1, "iter2": "a"}  # type: ignore
+    assert product_benchmark[0].fn() == (1, "a")
+    assert product_benchmark[1].fn.keywords == {"iter1": 1, "iter2": "b"}  # type: ignore
+    assert product_benchmark[1].fn() == (1, "b")
+    assert product_benchmark[2].fn.keywords == {"iter1": 2, "iter2": "a"}  # type: ignore
+    assert product_benchmark[2].fn() == (2, "a")
+    assert product_benchmark[3].fn.keywords == {"iter1": 2, "iter2": "b"}  # type: ignore
```
nicholasjng, line 56: Can you not go and check the benchmark interface here? Seems like it was made for this. (The fn having keywords available is equivalent to the interface having keywords that are not `inspect._empty`.)

tests/test_decorators.py
```diff
@@ -0,0 +1,65 @@
...
+    assert product_benchmark[3].fn() == (2, "b")
+
+
+def test_product_with_duplicate_parameters():
+    with pytest.warns(UserWarning, match="duplicate"):
+
+        @product(iter=[1, 1])
+        def product_benchmark(iter: int) -> int:
+            return iter
```
nicholasjng, line 65: Hah, I didn't know that, very cool. So for n = 1, `@nnbench.product == @nnbench.parametrize`.

tests/test_benchmarks_multidir_collection/benchmak_in_another_dir.py
```diff
@@ -0,0 +1,6 @@
+import nnbench
+
+
+@nnbench.benchmark(tags=("runner-collect",))
```
nicholasjng, line 4: File name is missing the "r" in benchmark.

tests/test_runner.py
```diff
@@ -1,18 +1,29 @@
...
     assert len(r.benchmarks) == 1
+    r.clear()
+
+    r.collect(testfolder, tags=("non-existing-tag",))
+    assert len(r.benchmarks) == 0
+    r.clear()
+
+    r.collect(os.path.join(testfolder), tags=("runner-collect",))
+    r.collect(os.path.join(another_testfolder), tags=("runner-collect",))
+    assert len(r.benchmarks) == 2
```
nicholasjng, line 26: I think you can `assert len(r.benchmarks) == 1` after the first collect and inline L16-L18 here.

tests/test_runner.py
```diff
@@ -31,3 +42,39 @@ def test_tag_selection(testfolder: str) -> None:
...
+
+    with pytest.raises(ValueError) as e:
+        r.run(
+            os.path.join(testfolder),
+            tags=("standard",),
+            params={"x": 1, "y": 1},
+            context=context_providers,
+        )
+
+    assert "got multiple values for context key 'system'" in str(e.value)
```
nicholasjng, line 80:
```suggestion
    with pytest.raises(ValueError, match="got multiple values for context key 'system'"):
        r.run(
            testfolder,
            tags=("standard",),
            params={"x": 1, "y": 1},
            context=context_providers,
        )
```

Any particular reason you're calling `os.path.join()` on a single directory all the time? For the trailing slash or?

nicholasjng (commented):
Looks great! Do let me know your thoughts about the state of the context provider - I was really annoyed by my first attempt when I couldn't just pass a plain `dict`.

Also, we should think about giving a high-level provider for the entire sysinfo struct - it's much better to just write e.g. `context=nnbench.context.sysinfo()` and get the whole platform info instead of passing a never-ending list.

(But that's for a follow-up.)

tests/test_runner.py
maxmynter: No, that was hwo the existing tests were implemented and I rolled with it. It's not needed for single dirs or files. Removing.

tests/test_runner.py
```diff
@@ -31,3 +39,37 @@ def test_tag_selection(testfolder: str) -> None:
...
+
+def test_error_on_duplicate_context_keys_in_runner(testfolder: str) -> None:
+    r = nnbench.BenchmarkRunner()
+
+    def duplicate_context_provider() -> dict[str, str]:
+        return {"system": "DuplicateSystem"}
+
+    context_providers = [system, duplicate_context_provider]
+
+    with pytest.raises(ValueError, match="got multiple values for context key 'system'") as e:
```
nicholasjng, line 69:
```suggestion
    with pytest.raises(ValueError, match="got multiple values for context key 'system'"):
```

nicholasjng (approved):
Nice!

PR #57, opened 2024-02-01
Add customization user guide

### LINKED ISSUE #35: Add user guide: Examples of customizing benchmarks and user-defined `nnbench.Parameters` for type safety
Add user guide to the documentation:
- [ ] Examples of customizing benchmarks (setup/teardown, context info, ...) and user-defined `nnbench.Parameters` for type safety

nicholasjng:
Covers setup/teardown tasks, context values and providers, and usage of `nnbench.Parameters`.

Introductory examples are given for every aspect to showcase how to get started easily with all of the different aspects.

----------------------------

[Return empty record on collection failure, add BenchmarkRunner to top-level module](https://github.com/aai-institute/nnbench/pull/57/commits/d53fee04d90c01fe004dfb40145a08bd183de61b)

This narrows the runner's return type, which will pay huge dividends for anyone building a custom reporter (which will soon be us).

Also, we now use `nnbench.BenchmarkRunner()` instead of importing from `nnbench.runner`.

Closes #35.

### DIFF
--- a/docs/guides/customization.md
+++ b/docs/guides/customization.md
@@ -1,0 +1,111 @@
+# Defining setup/teardown tasks, context, and `nnbench.Parameters`
+
+This page introduces some customization options for benchmark runs.
+These options can be helpful for tasks surrounding benchmark state management, such as automatic setup and cleanup, contextualizing results with context values, and defining typed parameters with the `nnbench.Parameters` class.
+
+## Defining setup and teardown tasks
+
+For some benchmarks, it is important to set certain configuration values and prepare the execution environment before running.
+To do this, you can pass a setup task to all of the nnbench decorators via the `setUp` keyword:
+
+```python
+import os
+
+import nnbench
+
+
+def set_envvar(**params):
+    os.environ["MY_ENV"] = "MY_VALUE"
+    
+
+@nnbench.benchmark(setUp=set_envvar)
+def prod(a: int, b: int) -> int:
+    return a * b
+```
+
+Similarly, to revert the environment state back to its previous form (or clean up any created resources), you can supply a finalization task with the `tearDown` keyword:
+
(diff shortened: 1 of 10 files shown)


docs/guides/customization.md
```diff
@@ -0,0 +1,109 @@
+# Defining setup/teardown tasks, context, and `nnbench.Parameters`
+
+This page introduces some customization options for benchmarks runs.
```
maxmynter, line 3:
```suggestion
This page introduces some customization options for benchmark runs.
```

docs/guides/customization.md
```diff
@@ -0,0 +1,109 @@
...
+For some benchmarks, it is important to set certain configuration values and prepare the execution environment before running.
+To do this, you can pass a setup task to all of the nnbench decorators via the `setUp` keyword:
+
+```python
+import os
+
+import nnbench
+
+
+def set_envvar(**params):
```
maxmynter, line 17: I would explicitly state that the `setUp` and `tearDown` functions must take all members of `params` as input. Otherwise it might not be clear why we pass `**params`. 

On that note, this is only stated in the `Benchmark` class docstring, not in the decorators. Should we add that there as well?

nicholasjng: Yes.

docs/guides/customization.md
nicholasjng: Fixed, thanks.

docs/guides/customization.md
nicholasjng: Added in [d56cac4](https://github.com/aai-institute/nnbench/pull/57/commits/d56cac46fb4729b991be0c96bd8e148ae3a36731), I will defer the decorator mention to a follow-up.

PR #58, opened 2024-02-02
Check type of parametrized argument to avoid mismatch

nicholasjng:
Newer test coverage has reminded us that we check argnames only, not argtypes in parameters of `@nnbench.parametrize` and `@nnbench.product`.

To fix, we grab the expected types from the signature and optionally unwrap them in the appropriate manner if they are generics or union types.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
        types_or: [ python, pyi ]
        args: [--ignore-missing-imports, --scripts-are-modules]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.14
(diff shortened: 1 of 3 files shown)


nicholasjng:
Repro:

```python
import nnbench
from nnbench.runner import BenchmarkRunner


@nnbench.parametrize(
    (dict(a=float(i), b=i+1) for i in range(2))
)
def strsum(a: str | int, b: str | int) -> str:
    return a + b


if __name__ == "__main__":
    runner = BenchmarkRunner()
    runner.run(__name__, params={})
```

prints `TypeError: expected type str | int, got type <class 'float'> for parametrized argument 'a'` on this branch.

PR #64, opened 2024-02-06
Add git info and Python package info providers

nicholasjng:
They are classes with minimal state and a no-arguments `__call__()` operator, giving a dictionary of context values as required.

Tests both providers in a lean unit test each.


nicholasjng:
It might make sense to move these semantically grouped context values down a notch, and use a top-level key for it:

So that, when e.g. using `GitEnvironmentInfo()`, the context goes from this (currently)

```python
r = nnbench.BenchmarkRunner()
res = r.run(..., context=(GitEnvironmentInfo(),))

res["context"]
# {"commit": "123456789abcdef", ...}
```

to this:

```python
res = r.run(..., context=(GitEnvironmentInfo(),))

res["context"]
# prints {"git": {"commit": "123456789abcdef", ...}}
```

This is possible within the current typing, since the value type is `Any`.

PR #71, opened 2024-02-08
Change `nnbench.reporter` file to submodule, add base reporter

nicholasjng:
The interface has been scratched in a GitHub issue previously, this is its (tentative) implementation.

It's not yet to be considered stable, and we will likely defer v0.2.0 until at least either a file or database reporter has been implemented.

Since the added benchmark record merge has raised the question of how we want to treat heterogeneous contexts, we should consider adding a proper Context object in a follow-up.

This is mostly a restructuring, so no new tests.


nicholasjng:
Questions (among others):

1) Do we keep the `_record` suffix if the input is named `record(s)`?
2) Do we keep `.report()` for backward compatibility (maybe only aliasing it to write())?

PR #73, opened 2024-02-08
Orchestrator examples

### LINKED ISSUE #67: Add example: Integration with workflow engines
Integration with workflow engines, structuring into a benchmarking task. (E.g. flyte, prefect, dagster, Beam, ...).

maxmynter:
Add Prefect example

To test it, you have to do the following:

- Install Prefect: `pip install -U prefect`
- Start prefect Server: `prefect server start`
- Expose the benchmark flows by executing the `runner.py` in the `examples/prefect` directory: `python runner.py`
- Then you can trigger an execution of the exposed flow by: `prefect deployment run 'train-and-benchmark/benchmark-runner'`

The default local UI of Prefect is `localhost:4200`.

Closes #67


examples/prefect/model_data_benchmark.py
```diff
@@ -0,0 +1,23 @@
+import pickle
```
nicholasjng, line 1: I would say merge the files into one - the tags provide acceptable semantic grouping already.

examples/prefect/runner.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def write(
+        self,
+        record: types.BenchmarkRecord,
+        benchmark_filter: str | None = None,
+        include_context: tuple[str, ...] = (),
+        exclude_empty: bool = True,
+    ) -> None:
+        old_stdout = sys.stdout
+        sys.stdout = StringIO()
```
nicholasjng, line 26: oje, you should never monkey-patch stdlib functions in code.

The way to go here is to attach a console handler to the obtained logger - see the documentation of the logging module on handlers: https://docs.python.org/3/library/logging.html#handler-objects

examples/prefect/training.py
```diff
@@ -0,0 +1,51 @@
...
+def make_train_test_split(
+    X: np.ndarray, y: np.ndarray, random_state: int, test_size: float = 0.2
+) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=test_size, random_state=random_state
+    )
+    return X_train, y_train, X_test, y_test
+
+
+@flow
```
nicholasjng, line 29: Do we ever need to prepare regression data outside of the flow afterwards? If not, we could think about inlining this into `prepare_regressor_and_test_data` below.

Is there any noticeable advantage of having it on the dashboard etc. in your opinion?

examples/prefect/training.py
```diff
@@ -0,0 +1,51 @@
...
+    X_train, y_train, X_test, y_test = prepare_regression_data(random_state=random_state)
+    model = train_linear_regression(X=X_train, y=y_train)
+    return model, X_test, y_test
+
+
+@task
+def train_linear_regression(X: np.ndarray, y: np.ndarray) -> base.BaseEstimator:
+    model = LinearRegression()
+    model.fit(X, y)
+    return model
```
nicholasjng, line 51: Another candidate for inlining. (Also please define all tasks after the flows, seems to be a convention for Python orchestrators in general.)

examples/prefect/model_data_benchmark.py
```diff
@@ -0,0 +1,23 @@
...
+    return sys.getsizeof(model_bytes)
+
+
+@nnbench.benchmark(name="Inference time (s)", tags=("model-meta",))
+def inference_time(model: base.BaseEstimator, X: np.ndarray, n_iter: int = 100) -> float:
+    start = time.perf_counter()
+    for i in range(n_iter):
+        _ = model.predict(X)
+    end = time.perf_counter()
+    return (end - start) / n_iter
```
nicholasjng, line 23: We time each benchmark anyway, so we technically have this in the bag. Any idea how could we give the option to use the execution time as the target metric instead?

(Otherwise, it's fine - though the `time_ns` metric then becomes superfluous.)

examples/prefect/model_data_benchmark.py
```diff
@@ -0,0 +1,23 @@
...
+import numpy as np
+from sklearn import base
+
+import nnbench
+
+
+@nnbench.benchmark(name="Model size (bytes)", tags=("model-meta",))
+def modelsize(model: base.BaseEstimator) -> int:
+    model_bytes = pickle.dumps(model)
+    return sys.getsizeof(model_bytes)
```
nicholasjng, line 14: Is there any reason to not just call `sys.getsizeof()` on the model directly?

examples/prefect/benchmark.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+from prefect import task
+from sklearn import base, metrics
+
+import nnbench
+
+
+@nnbench.benchmark
```
nicholasjng, line 8: I would leave this file out until we can think of a way to facilitate the nnbench.Benchmark<->prefect.Task correspondence.

examples/prefect/model_data_benchmark.py
maxmynter: Calling on the object directly does only give the memory footprint of the object itself, not of any other values it references. See [here](https://stackoverflow.com/questions/19516403/why-pickle-dumpobj-has-different-size-with-sys-getsizeofobj-how-to-save-var). Thats why I went with serializing first.

examples/prefect/model_data_benchmark.py
maxmynter: Good point. I'll think of another metadata metric we can use.

examples/prefect/benchmark.py
maxmynter: Damn, yeah. I deleted it but it must have somehow come back  while squashing when i was cleaning up the git history.

examples/prefect/training.py
maxmynter: I was kinda on the fence about this one and had it inline at first. 
It makes a nice dependency graph in the Prefect UI but we are also not demo'ing prefect and the difference is not that big. Factoring it back

examples/prefect/training.py
maxmynter: The thing is if we inline everything the example may become too small to show the integration into Prefect.

And you mean flows after tasks, right? At least thats what the Prefect docs do.

examples/prefect/runner.py
maxmynter: Mhhh that does not seem to work. Prefect supplies a `LoggerAdapter` not a `Logger`. To the former one cannot attach streams.

maxmynter: Decided on leaving out formatted reporting. In a real life scenario reports would go into a database and would then be read from a dedicated UI, not the orchestrator. 

I will implement a barebones reporter that simply logs the `dict` returned by the benchmarks to the console though.

examples/prefect/model_data_benchmark.py
maxmynter: Though the `time_ns ` times the whole benchmark. Here we know the time it takes for the plain inference. Both are different (esp. with more complex benchmarks) but of interest, no?

nicholasjng: Yeah, what I meant is that both should be the same if the benchmark consists only of "_ = model.predict(X)".

But that's statistically unstable, so please leave it as is.

examples/prefect/training.py
nicholasjng: Yes, flows after tasks, sorry.

docs/tutorials/prefect.md
```diff
@@ -0,0 +1,90 @@
+# Integrating nnbench with Prefect
+
+If you have more complex workflows it is sensible to use a workflow orchestration tools. 
+Benchmarking with nnbench can be integrated with orchestrators. We will present an example integration with Prefect.
+We will explain the orchestration concepts in a high level and link to the corresponding parts of the  Prefect [docs](https://docs.prefect.io/).
+The full example code is in the nnbench [repository](https://github.com/aai-institute/nnbench/tree/main/examples/prefec).
```
nicholasjng, line 6:
```suggestion
The full example code can be found in the nnbench [repository](https://github.com/aai-institute/nnbench/tree/main/examples/prefect).
```

docs/tutorials/prefect.md
```diff
@@ -0,0 +1,90 @@
+# Integrating nnbench with Prefect
+
+If you have more complex workflows it is sensible to use a workflow orchestration tools.
```
nicholasjng, line 3:
```suggestion
If you have more complex workflows, it is sensible to use a workflow orchestration tool to manage them. 
```

docs/tutorials/prefect.md
```diff
@@ -0,0 +1,90 @@
...
+## Project Structure
+### Defining the training tasks and workflows
+We recommend to separate the training and benchmarking logic. 
+
+```python
+--8<-- "examples/prefect/training.py"
+```
+
+The `training.py` file contains functions to generate synthetic data for our regression model, facilitate a train-test-split, and finally train the regression model.
+We have applied Prefects [`@task` decorator.](https://docs.prefect.io/latest/concepts/tasks/)
```
nicholasjng, line 19:
```suggestion
We have applied Prefect's [`@task` decorator.](https://docs.prefect.io/latest/concepts/tasks/)  
```

I would also merge the next sentence into it, e.g. like "which marks the function logic ..."

docs/tutorials/prefect.md
```diff
@@ -0,0 +1,90 @@
...
+
+```python
+--8<-- "examples/prefect/training.py"
+```
+
+The `training.py` file contains functions to generate synthetic data for our regression model, facilitate a train-test-split, and finally train the regression model.
+We have applied Prefects [`@task` decorator.](https://docs.prefect.io/latest/concepts/tasks/)  
+It marks the contained logic as a discrete unit of work for Prefect. 
+Two other functions prepare the regression data and train the estimator. 
+They are labeled with the [`@flow` decorator.](https://docs.prefect.io/latest/concepts/flows) which labels the function as a workflow logic container that can depend on other flows or tasks.
```
nicholasjng, line 22:
```suggestion
They are labeled with the [`@flow` decorator.](https://docs.prefect.io/latest/concepts/flows) which labels the function as a workflow that can depend on other flows or tasks.
```

docs/tutorials/prefect.md
```diff
@@ -0,0 +1,90 @@
...
+The benchmarks are in the `benchmark.py` file.
+We have two functions to calculate the mean absolute error and the mean squared error.
+These benchmarks are tagged to indicate they are metrics.
+The other two benchmarks calculate model metadata, namely the size and inference time of the model.
+
+```python
+--8<-- "examples/prefect/benchmark.py"
+```
+
+We did not apply any Prefect decorators here as they are incompatible with nnbench's decorators.
```
nicholasjng, line 35: True, but they are also a finer-grained unit of work, as we currently operate under the "one benchmark family per task" paradigm. I would mention that here instead.

nicholasjng (commented):
Sweet!

Issue #48, opened 2024-02-01
Add flexible base reporters for file IO, web, and DBs


nicholasjng:
Each of these should inherit directly from `BenchmarkReporter` and define a nice interface for the respective IO to be reusable or even useable out of the box for most users.

Pseudocode example:

```py
class FileIOReporter:
    def write_record(self, r):
        ...

    def write_record_batched(self, rb):
        ...

    def read_record(self):
        ...
    
    def open(self, fp):
        ...

    def close(self, fp):
        ...


class JSONFileReporter(FileIOReporter):

    def open(self, fp):
        with open(fp, "r") as f:
            return json.load(f)

    def close(self, fp):
        fp.close()


class YAMLFileReporter(FileIOReporter):
    ... # same thing as for JSON, but use YAML read/write APIs.
```

Objective: Implement the base BenchmarkReporter's interface for a few specific families of reporters, not every reporter by itself.

Bonus: Maybe the IO (open, close) can even become a mixin that makes specific file or database reporters trivial by just acquiring and releasing a handle to write to.

Hrsj123:
I was wondering what is the advantage of having def read_record() method? Does it have any specific use?

Hrsj123:
Thanks for your clarification, it makes sense.

PR #74, opened 2024-02-08
Added IOMixin for reporter

Hrsj123:
issue #48 

I've tried to create a new IOMixin class that provides methods for opening, reading, writing, and closing files at a given path.

Please review the code and share your feedback on it.

### DIFF
--- a/src/nnbench/reporter/mixins.py
+++ b/src/nnbench/reporter/mixins.py
@@ -1,0 +1,45 @@
+import os
+from typing import Any
+
+
+class IOMixin:
+    """
+    A mixin class providing file I/O operations for benchmark reporters.
+
+    This mixin offers common file handling functionalities like opening, closing, reading and writing.
+    This class can be used to create any BenchmarkReporter which has requirement of reading file.
+    """
+
+    def open(self, file_path: str) -> None:
+        """
+        Opens a existing file at a given file_path for read and write operations.
+
+        Parameters
+        ----------
+        file_path : str
+            The path to the target file or directory.
+
+        Raises
+        ------
+        FileNotFoundError
+            If the target file is not present at the file_path.
+        """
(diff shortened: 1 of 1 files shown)


nicholasjng:
Thanks for the PR. I assume you crafted this from my suggestion in #48?

We rearranged some APIs on the tentative base reporter interface, which I merged yesterday in https://github.com/aai-institute/nnbench/pull/71.

There, instead of `open()` and `close()`, we have an API set consisting of `initialize/finalize` (to acquire/release resources that are needed by a reporter during its object lifetime), `read/write` (for benchmark record IO), and batched versions thereof.

We decided to (temporarily) not go ahead with `open()` and `close()` because they are too specific to the file case. 
FWIW, the [Beam file-based sink](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/filebasedsink.py#L45), which was our main inspiration for the API set presented in #48, is directly tied to file IO as well.

That's not to say that this is useless, though. I'm wondering how we could use such an abstraction to abstractify file writes to whichever format we like, such as JSON, YAML, TOML, the usual suspects.

If you'd like to get involved in that, I would suggest the following:
1) Pull in current `main`, and create a feature branch for a file reporter.
2) Add a Python file in `src/nnbench/reporter/`, maybe `file.py`, which holds the file reporter.
3) The constructor should take a result directory, which we save the records to.
4) The `initialize()` method should create this directory, if not already existent.
5) The `read()` method should open a single file in that directory, and parse the data with the given format, like e.g. JSON, TOML, YAML.
6) The `write()` method should write the record to a file in that directory, just like you implemented here.
7) Finalize can be empty, or optionally destroy the directory again if you'd like.

Once we have that, we can think about how to make this work across file formats. How does that sound?

Hrsj123:
I got your point, I will try creating a generic file reporter.

PR #76, opened 2024-02-12
Inline console reporter into BenchmarkReporter.display()

nicholasjng:
Since the ability to print to console is so nice, we change the console reporter to be available on every reporter descendant by adding it as the `display()` method on the base.

This means that a) the console reporter file can go, and b) tabulate has to be upgraded to an obligatory dependency (at least until we have a better tabular formatter in rich etc.).

-------------------

[Remove console reporter file, add tabulate to project requirements](https://github.com/aai-institute/nnbench/commit/2f57cc211a4fef873e52bb5927d002e321df5cc0)

Also updates the lock files, although not much should change because it was a dev dependency already.

Removes the console reporter from the registry, which is unused at this point, and might be ripe for removal until we add a CLI running facility.

Updates all the docs to use the base `BenchmarkReporter()` and its `display()` method.


nicholasjng:
fyi: I've also taken out the "merge records for batched writing" strategy because a) it was introduced in the first place to write multiple records to console, and b) it doesn't work too well for duckdb, indicating that the strategy is not refined enough.

PR #78, opened 2024-02-12
Feature/file reporter

Hrsj123:
Closes #74

Here, I have written a generic FileReporter according to your recommendations mentioned in PR #74.
To specialize the method to both reading and writing to file in any particular style, we can utilize the Parser class to define read and write methods.
I have defined JSONParser and YAMLFileParser as sample in `file.py` for handling read and write operations in json and yaml files.
Since the FileReporter read and write functions utilized the Parser class, it automatically adjusts the read and write methods based on the file extension we are writing to.

### Registering a Parser
```
class MyCustomParser(Parser):
        def parse_file(self, records):
            # Implement your custom parsing logic here
            ...
        def write_records(self, records):
            # Implement your custom file writing logic here
            ...
# Register your custom parser with a distinct file type
MyCustomParser.register("my_custom_format")
```
Once the parser is registered with a particular extension, we can utilize that parser implicitly based on the extension of `file_name` we want to write to.

Do let me know your views and recommendations on these changes.

### DIFF
--- a/src/nnbench/reporter/file.py
+++ b/src/nnbench/reporter/file.py
@@ -1,87 +1,132 @@
 from __future__ import annotations
 
 import os
+from typing import Any, List
 
 from nnbench.reporter.base import BenchmarkReporter
 from nnbench.types import BenchmarkRecord
 
-from abc import ABC, abstractmethod
 
+class Parser:
+    """The base interface for parsing records form file.
 
-class Parser(ABC):
-    """Abstract base class for parsing records form file.
-    
     Usage:
     ------
     ```
     class MyCustomParser(Parser):
-        def parse(self, records):
+        def parse_file(self, records):
             # Implement your custom parsing logic here
+            ...
+        def write_records(self, records):
+            # Implement your custom file writing logic here
             ...
     # Register your custom parser with a distinct file type
     MyCustomParser.register("my_custom_format")
     # Usage:
-    custom_records = ...  # Load records in your custom format
-    append_record_to_records(records, custom_record, "my_custom_format")
+    new_record = ...  # Load records in your custom format
+    append_record_to_records(records, new_record, "my_custom_format")
     ```
     """
 
-    @abstractmethod
-    def parse(self, records):
+    def parse_file(self, records: str) -> Any:
         """Parses records and returns a list of parsed data.
 
         Args:
-            records: A list or iterator of record strings.
+        -----
+            `records:` A list or iterator of record strings.
 
         Returns:
+        --------
             A list of parsed records.
         """
+        raise NotImplementedError
+
+    def write_records(self, records: Any[BenchmarkRecord], record: BenchmarkRecord) -> str:
+        """Appends a record to the existing records based on the file type.
+
+        Args:
+        -----
+            `records:` A list of parsed records.
+            `record:` The record string to append.
+            `file_type:` The file type (string).
+
+        Returns:
+        --------
+            A string form of the content to be written in a file.
+        """
+        raise NotImplementedError
 
     @classmethod
-    def register(cls, file_type):
+    def register(cls, file_type: str) -> None:
         """Registers a parser for a specific file type.
 
         Args:
-            file_type: The file type (string)
+            `file_type:` The file type (string)
         """
         parsers[file_type] = cls
 
     @staticmethod
-    def get_parser(file_type):
+    def get_parser(file_type: str):
         """Gets the registered parser for a file type.
 
         Args:
-            file_type: The file type (string)
+            `file_type:` The file type (string)
 
         Returns:
+        --------
             The registered RecordParser, or None if not found.
         """
+        return parsers.get(file_type)
 
 
 class JsonParser(Parser):
-    def parse(self, records):
+    def parse_file(self, records: str) -> List[dict]:
         import json
-        return [json.loads(record) for record in records]
+
+        try:
+            return json.loads(records if records else "[]")
+        except json.JSONDecodeError:
+            raise ValueError("Unexpected records passed")
+
+    def write_records(
+        self, parsed_records: Any[BenchmarkRecord] | None, record: BenchmarkRecord
+    ) -> str:
+        import json
+
+        parsed_records.append(record)
+        return json.dumps(parsed_records)
 
 
 class YamlParser(Parser):
-    def parse(self, records):
+    def parse_file(self, records: str) -> List[dict]:
         import yaml
-        return [yaml.safe_load(record) for record in records]
+
+        return yaml.safe_load(records) if records else []
+
+    def write_records(
+        self, parsed_records: Any[BenchmarkRecord] | None, record: BenchmarkRecord
+    ) -> str:
+        import yaml
+
+        parsed_records.append(record)
+        for element in record["benchmarks"]:
+            element["value"] = float(element["value"])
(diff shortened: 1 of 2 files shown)


src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
+from __future__ import annotations
+
+import os
+from typing import Any, List
+
+from nnbench.reporter.base import BenchmarkReporter
+from nnbench.types import BenchmarkRecord
+
+
+class Parser:
```
nicholasjng, line 10: Just for clarification: We're not actually parsing things when we are reading files, that is the job of the respective modules (`json|yaml|toml`).

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+        import yaml
+
+        parsed_records.append(record)
+        for element in record["benchmarks"]:
+            element["value"] = float(element["value"])
+        return yaml.dump(parsed_records)
+
+
+# Register custom parsers here
+parsers = {"json": JsonParser, "yaml": YamlParser}
```
nicholasjng, line 118: I like the idea of making it a module-wide default, but we should take care that

a) the variable is private (i.e. has a leading underscore) to prevent accidental export, and 
b) we should make the value structure of the map as easy as possible.

To address b), I would start by making it a `tuple[ser, de]` where `ser` is a `Callable[[IO, dict[str, Any]], None]`, i.e., a function taking a file descriptor in write mode and a record and writing it to a file, and `de` being a `Callable[[IO], dict[str, Any]]`, a function taking a file descriptor in read mode and returning the loaded record.

You can then register the SerDe factories based on whether the necessary packages are installed (json is available out of the box, yaml and toml are not).

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+    parser = Parser.get_parser(file_type)
+    if parser is None:
+        raise ValueError(f"Unsupported file type: {file_type}")
+
+    return parser().write_records(parsed_records, record)
+
+
+class FileReporter(BenchmarkReporter):
+    def __init__(self, dir: str):
+        self.dir = dir
```
nicholasjng, line 163: `dir` is a builtin name, so I would change this to e.g. `directory` or `result_dir`.

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+        if not os.path.exists(dir):
+            self.initialize()
+
+    def initialize(self) -> None:
+        try:
+            os.makedirs(self.dir, exist_ok=True)
+        except OSError as e:
+            self.finalize()
+            raise ValueError(f"Could not create directory: {self.dir}") from e
```
nicholasjng, line 172: You can just call `Path(self.dir).mkdir(exist_ok=True, parents=...)` - if that errors, you might not have to finalize anything.

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+            raise BaseException("Directory is not initialized")
+        file_path = os.path.join(self.dir, file_name)
+        file_type = file_name.split(".")[1]
+        try:
+            with open(file_path) as file:
+                data = file.read()
+                parsed_data = parse_records(data, file_type)
+                return parsed_data
+        except FileNotFoundError:
+            raise ValueError(f"Could not read the file: {file_path}")
```
nicholasjng, line 185: This needs the restructured file loading dict I talked about earlier, but in essence, all that should happen here is the file being loaded with `open` (like you did), and then calling the deserializer on the opened file.

In particular, no error handling for `open()` should be necessary, since those are informative enough for the user on their own.

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+            with open(file_path, "w") as file:
+                file.write("")
+        try:
+            parsed_records = self.read(file_name)
+            file_type = file_name.split(".")[1]
+            new_records = append_record_to_records(parsed_records, record, file_type)
+            with open(file_path, "w") as file:
+                file.write(new_records)
+        except FileNotFoundError:
+            raise ValueError(f"Could not read the file: {file_path}")
```
nicholasjng, line 202: Same here, just load the serializer from the dict and call it on the opened file.

src/nnbench/reporter/file.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def finalize(self) -> None:
+        del self.dir
```
nicholasjng, line 205: This does not do what you think it does - it just stages the `self.dir` variable for garbage collection. 
To safely remove the directory (which you might not want to do anyway, we could add a flag in the constructor for that?), you should call `shutil.rmtree(self.dir, ignore_errors=True)`. (Though you might want to check existence first and set ignore_errors to `False`.

nicholasjng (commented):
Thanks - this looks promising, but we'll have to tweak it slightly to make it as effective as possible.

If anything is unclear, either ping me here or I can push some aspects as I have them in my head to your branch if you prefer.

src/nnbench/reporter/file.py
Hrsj123: Should I use a class-based approach to register the SerDe factories (like I used in Parser class) ? 
One other option is to use a simple register method which takes 3 arguments (i.e., Ser and De functions and a file_type)  as arguments.
Also, I think the class-based approach is much more concise to define the ser and de methods on the user side.

nicholasjng: I'm happy with either, though the optional import part (i.e., erroring if `toml/yaml` are not installed) will be a bit easier in the class case.

For now, I think the quickest way is a functional approach, though. Maybe like this:

```python
_file_loaders: dict[str, tuple[Any, Any]] = {}

def yaml_load(fp: IO, options=None):
    try:
        import yaml
    except ImportError:
        raise ModuleNotFoundError("`pyyaml` is not installed")
    
    # takes no options, but the slot is useful for passing options to file loaders.
    obj = yaml.safe_load(fp)
    return BenchmarkRecord(context=obj["context"], benchmarks=obj["benchmarks"])

def yaml_save(record: BenchmarkRecord, fp: IO, options=None) -> None:
    try:
        import yaml
    except ImportError:
        raise ModuleNotFoundError("`pyyaml` is not installed")

    yaml.safe_dump(record, fp, **(options or {})


_file_loaders["yaml"] = (yaml_save, yaml_load)
```

With an option of defining e.g. a `register_file_io(ser, de)` later to do the dict insertion if we want.

Hrsj123:
I've changed the value value of the `_file_loader` registry and made the necessary changes recommended by you. Do review the changes and let me know if anything else can be improved.

nicholasjng:
Thanks!

To get you up to speed to the current situation, we (I) have found that a file reporter is necessary as a building block for the duckDB reporter implementation that I started in #75. For this reason, I implemented a file reporter myself in that PR, which we will go ahead with ASAP.

That is not meant as a knock on you - you did a great job here, and a lot of the logic you brought in here also made it into #75, we just had to speed things up a bit to consolidate the interface. If you are interested, I'd like to welcome you to help improve on it (especially the file driver registration/deregistration hooks you gave here come to mind)!

I'm going to open some tickets related to improvement of that file IO solution, which I can assign to you if you want, just let me know by commenting on the respective issue.

PR #79, opened 2024-02-12
Add BenchmarkContext Class

### LINKED ISSUE #72: Add proper `Context` object, manipulation methods
We've so far rolled pretty good with a bare-dict context, but it might be time to change that.

Newly, we encountered the problem of merging two benchmark records. An easy first solution is inlining the context into the accompanying benchmark list, but that has the problem of not being idempotent.

To fix, we could think of merging two context dictionaries by taking the union of the keys and padding missing values with a unique placeholder.

Together with the previously introduced filtering, flattening, and nested getitem methods, this points towards a custom class having value over a bare dictionary.

maxmynter:
Closes #72

## Todo
- [x] Integrate Context logic into runner
- [x] and reporter

### DIFF
--- a/src/nnbench/context.py
+++ b/src/nnbench/context.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import itertools
 import platform
 import sys
-from typing import Any, Callable, ItemsView, KeysView, Literal, ValuesView
+from typing import Any, Callable, Iterator, Literal
 
 ContextProvider = Callable[[], dict[str, Any]]
 """A function providing a dictionary of context values."""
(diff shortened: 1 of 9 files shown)


src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+        Merge another dictionary, Context, or ContextProvider into this Context.
+
+        Parameters
+        ----------
+        other : Context | dict[str, Any] | ContextProvider
+            The other context to merge into this one.
+        this_ctx : dict[str, Any] | None, optional
+            The target dictionary to merge into. Takes this classes context if None.
+        inplace : bool, optional
+            If True, modifies the current context in place. Otherwise, returns a new merged Context.
```
nicholasjng, line 223: We need to be super careful with this to avoid surprises.

Returning a different instance from an instance method seems odd. How does pandas do this? A different instance with a view on the original?

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+    ) -> "Context":
+        """
+        Merge another dictionary, Context, or ContextProvider into this Context.
+
+        Parameters
+        ----------
+        other : Context | dict[str, Any] | ContextProvider
+            The other context to merge into this one.
+        this_ctx : dict[str, Any] | None, optional
+            The target dictionary to merge into. Takes this classes context if None.
```
nicholasjng, line 221: This is redundant with `self` in OOP terms, so please leave it out.

(Alternatively: You probably would not want to merge two things in a class into a different object, that's functional programming.)

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+                raise ValueError(f"Key collision upon merge, {key}")
+
+        if inplace:
+            self._ctx_dict = merged
+            return self
+        else:
+            return Context(context=merged)
+
+    @staticmethod
+    def flatten_dict(d: dict[str, Any], prefix: str = "", sep: str = ".") -> dict[str, Any]:
```
nicholasjng, line 275: This should probably be private, since it operates on a private member.

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+        """
+
+        if inplace:
+            self._ctx_dict = self.flatten_dict(d=self._ctx_dict, prefix=prefix, sep=sep)
+            return self
+        else:
+            return Context(context=self.flatten_dict(self._ctx_dict, prefix=prefix, sep=sep))
+
+    @staticmethod
+    def unflatten_dict(d: dict[str, Any], sep: str = ".") -> dict[str, Any]:
```
nicholasjng, line 329: Same as above, this should probably be private.

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+            The current context instance if `inplace` is True; otherwise, a new unflattened Context instance.
+        """
+        if inplace:
+            self._ctx_dict = self.unflatten_dict(self._ctx_dict, sep=sep)
+            return self
+        else:
+            return Context(context=self.unflatten_dict(self._ctx_dict, sep=sep))
+
+    @staticmethod
+    def filter_dict(d: dict[str, Any], predicate: Callable[[str, Any], bool]) -> dict[str, Any]:
```
nicholasjng, line 379: And again, though I think this might be a deletion candidate once we implement nested key/value views.

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
         result["memory_unit"] = self.memunit
         # TODO: Lacks CPU cache info, which requires a solution other than psutil.
         return {self.key: result}
+
+
+class Context:
+    def __init__(self, context: dict[str, Any] | ContextProvider | "Context" | None = None) -> None:
```
nicholasjng, line 193: Unfortunate name, I think this should be `_data` or `data` or something.

src/nnbench/context.py
```diff
@@ -183,3 +187,256 @@ def __call__(self) -> dict[str, Any]:
...
+            other = other._ctx_dict
+        elif callable(other):
+            result = other()
+            if not isinstance(result, dict):
+                raise ValueError("Provider did not return a dictionary.")
+            other = result
+        elif isinstance(other, dict):
+            pass
+        else:
+            raise TypeError(f"Unknown type for source, got {type(other)}")
```
nicholasjng, line 248: I think all of this would be dramatically easier if we operated on a list of items.

Proposition (as stated above): Instead of merge(), we could support an update() method. We lose the immutability of the left-hand side context, but I think that operation is much more common.

tests/test_context.py
```diff
@@ -39,3 +61,43 @@ def test_cpu_info_provider() -> None:
...
+
+def test_flatten_nested_dictionary(nested_context_instance):
+    flattened = nested_context_instance.flatten(inplace=False)
+    assert flattened._ctx_dict == flat_dict
+
+
+def test_flatten_with_prefix(nested_context_instance):
+    flattened = nested_context_instance.flatten(prefix="prefix", inplace=False)
+    expected_dict = {f"prefix.{k}": v for k, v in flat_dict.items()}
+    assert flattened._ctx_dict == expected_dict
```
nicholasjng, line 85: Does this work with a nested dict, too?

tests/test_context.py
```diff
@@ -1,4 +1,26 @@
...
+
+
+@pytest.fixture
+def flat_context_instance():
+    return Context(flat_dict)
+
+
+@pytest.fixture
+def nested_context_instance():
+    return Context(nested_dict)
```
nicholasjng, line 23: I think you can provision them as necessary, if anything, the data above this should be a fixture.

nicholasjng (commented):
This is big! We should keep in mind to keep everything nicely serializable, but that should be easy (it's just a fancy dict wrapper in the end).

Could you / we come up with examples of how to achieve the following with our records when writing to disk / DB:

1. Write normally (unflattened everything, only two top-level keys "context" and "benchmarks")
2. Context inlined, but whole (i.e., as a "context" key in all benchmark JSON objects)
3. Context inlined and broadcasted/flattened, i.e. all keys of the flattened context are in the benchmark JSON objects)

Let's also think of reversibility of such actions (can I reverse number 3) if I explode the context into the benchmark struct? Probably not right?).

Once we have that, we need to add those options to the serde factory in the duckDB PR.

src/nnbench/context.py
maxmynter: Pandas returns None for actions performed inplace (`inplace=True`) and a copy of the object with the action performed on `inplace=False`. I would go with their design to remain consistent. Not returning `self` also prohibits reassigning on non inplace operations. Thus there is a visual distinction between the two ways when reading code.

Also they have `inplace=False` as a default consistently to prevent unintentional data modifications. People won't work in notebooks with nnbench so the risk of changeing state in a session is lower. Still, I would go with the pandas way to make inplace changes deliberate and to use a familiar API design pattern.

src/nnbench/context.py
maxmynter: We need the ability to pass a dict for the recursive call later. I am taking the same approach here as with the other methods (a private method working on dicts and an exposed instance method collectin the context dicts and passing them to the private one).

src/nnbench/context.py
```diff
@@ -1,12 +1,16 @@
...
 import platform
 import sys
-from typing import Any, Callable, Literal
+from typing import Any, Callable, Iterator, Literal
 
 ContextProvider = Callable[[], dict[str, Any]]
 """A function providing a dictionary of context values."""
 
+ContextElement = dict[str, Any] | ContextProvider
+
```
nicholasjng, line 13: You don't appear to be using this anywhere, so I guess it can go.

src/nnbench/context.py
```diff
@@ -183,3 +187,158 @@ def __call__(self) -> dict[str, Any]:
...
+    def items(self, sep: str = ".") -> Iterator[tuple[str, Any]]:
+        yield from self._ctx_items(d=self._ctx_dict, prefix="", sep=sep)
+
+    def update(self, other: ContextProvider | dict[str, Any] | "Context") -> None:
+        if callable(other):
+            other = other()
+        elif isinstance(other, Context):
+            other = other._ctx_dict
+        self._ctx_dict.update(other)
```
nicholasjng, line 223: This means that only top-level key updates will work. But what about an update like 

```python
d = {"a": 1, "b": 2, "c": {"d": 3}}; d.update({"c.d": 4})
```

?

src/nnbench/reporter/base.py
```diff
@@ -114,7 +115,7 @@ def display(
                 continue
             filteredctx = {
                 k: v
-                for k, v in flatten(ctx).items()
+                for k, v in Context._flatten_dict(ctx).items()
```
nicholasjng, line 118: I think we should `Context.make(bm["context"])` early, and then filter on it specifically. We can implement this together if you like.

tests/test_context.py
```diff
@@ -39,3 +39,22 @@ def test_cpu_info_provider() -> None:
     assert res["frequency"] > 0
     assert res["num_cpus"] > 0
     assert res["total_memory"] > 0
+
+
+def test_flatten_nested_dictionary():
+    nested_ctx = Context({"a": 1, "b": {"c": 2, "d": {"e": 3}}})
+    flattened = nested_ctx.flatten(inplace=False)
+    assert flattened._ctx_dict == {"a": 1, "b.c": 2, "b.d.e": 3}
```
nicholasjng, line 47: I'd love to see some `keys|values|items` tests, specifically for their typing and to see that they behave expectedly.

src/nnbench/reporter/base.py
maxmynter: Happy to do that!

nicholasjng (approved):

PR #84, opened 2024-02-21
Add compression algorithms facility to file IO 

nicholasjng:
~Still work in progress, I'm currently debating which other compressions to add (lzma? xz? ...), and whether to expose the stateful hooks (`get_driver_implementation() | get_compression_algorithm()`) right now instead of letting people roll their own SerDe registry.~

------------------

Works similarly to the file driver lookup, except the compression can
also be None, in which case a normal file descriptor is used.

Canonicalizes the extension of the resulting file regardless of which
file name is used based on driver and compression algorithm. This might
be surprising behavior, but simplifies driver and compression inference
significantly on roundtrips.

Scraps the custom registries in FileIO constructor, because we intend to
expose reg/dereg hooks for the default later on.

Available compressions are gzip and bz2, which are cross-platform Python stdlib builtins.

### DIFF
--- a/src/nnbench/reporter/file.py
+++ b/src/nnbench/reporter/file.py
@@ -4,7 +4,7 @@ import os
 import threading
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import IO, Any, Callable, Literal, Sequence
+from typing import IO, Any, Callable, Literal, Sequence, cast
 
 from nnbench.reporter.base import BenchmarkReporter
 from nnbench.types import BenchmarkRecord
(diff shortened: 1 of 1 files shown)


src/nnbench/reporter/file.py
```diff
@@ -113,33 +113,63 @@ def csv_load(fp: IO, fdoptions: FileDriverOptions) -> list[BenchmarkRecord]:
...
+with _compression_lock:
+    _compressions["gzip"] = gzip_compression
+    _compressions["bz2"] = bz2_compression
+
+
+def get_compression_algorithm(name: str) -> Callable:
+    try:
+        return _compressions[name]
+    except KeyError:
+        raise KeyError(f"unsupported compression algorithm {name!r}") from None
```
maxmynter, line 146: Question: Why raise from `None` and not a plain raise?

maxmynter (approved):

src/nnbench/reporter/file.py
nicholasjng: `raise from None` cuts off the parent exception, which is the KeyError in the `_compressions` dict lookup. That does not give you any new information, so I decided to throw it away.

PR #85, opened 2024-02-22
Add BQ example to showcase cloud streaming of benchmark records

### LINKED ISSUE #66: Add example: Streaming results to a database
Add example or tutorial to the documentation on streaming results to a database with a custom data sink.
Please note: https://duckdb.org/2023/03/03/json.html for getting started with DuckDB.

- [x] Prepare BQ example
- [ ] Check/verify SQL model
- [ ] Decide on what API we go forward with

nicholasjng:
In this setup, we show how easy it is to stream records completely
without a reporter by simply ingesting the JSON representations resulting
from `BenchmarkRecord.compact()`.

This might be the way forward for most of our examples in place of a
dedicated reporter.

---------------

[Clean up mypy config to allow dupe filenames in examples/](https://github.com/aai-institute/nnbench/commit/29b38cbc331f57d67a0935e37921edac4898ebd3)

< no message>

[Make contextkeys() an API to handle separator properly](https://github.com/aai-institute/nnbench/commit/a13b7f698dd08e25d035ff22a84b3b784d301b90)

The BQ example revealed that we cannot just store the keys as a property
on our class, since they might not be digestible by a database if certain
illegal characters are used.

For BQ, this was the case with the default "." separator, which means we
have to pipe the separator through in the `BenchmarkRecord.compact()`
call. Hence, the context keys need to be fetched dynamically.

Closes #66.

### DIFF
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -22,7 +22,9 @@ class BenchmarkRecord:
     benchmarks: list[dict[str, Any]]
 
     def compact(
-        self, mode: Literal["flatten", "inline", "omit"] = "inline"
+        self,
+        mode: Literal["flatten", "inline", "omit"] = "inline",
+        sep: str = ".",
     ) -> list[dict[str, Any]]:
         """
         Prepare the benchmark results, optionally inlining the context either as a
@@ -34,6 +36,8 @@ class BenchmarkRecord:
             How to handle the context. ``"omit"`` leaves out the context entirely, ``"inline"``
             inserts it into the benchmark dictionary as a single entry named ``"context"``, and
             ``"flatten"`` inserts the flattened context values into the dictionary.
+        sep: str
+            The separator to use when flattening the context, i.e. when ``mode = "flatten"``.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


src/nnbench/types.py
```diff
@@ -20,13 +20,14 @@ def NoOp(**kwargs: Any) -> None:
 class BenchmarkRecord:
     context: Context
     benchmarks: list[dict[str, Any]]
-    _ctx_keys: tuple[str, ...] = field(init=False, repr=False)
 
-    def __post_init__(self):
-        super().__setattr__("_ctx_keys", tuple(self.context.keys()))
+    def contextkeys(self, sep: str = ".") -> tuple[str, ...]:
+        return tuple(self.context.keys(sep=sep))
```
maxmynter, line 25: I just removed this in the Streamlit PR following your suggestion. But i prefer it this way. It's also necessary as we need to pass the `sep` for BigQuery, right? 

Let's merge this one before the Streamlit example and I'll make this change supercede the deletion and test for remaining issues with Prefect/Streamlit or merge conflicts. 

Anything else, I should know of w.r.t. to `contextkeys()`

maxmynter (approved):
Nice

PR #91, opened 2024-02-26
Add duckDB example

nicholasjng:
Showcases how to query file-written records obtained from a local benchmark run, both with an inlined context and a flattened one.

Contains a SQL query example with escaped dotted columns, which are necessary as we currently cannot specify the separator to use when writing files to disk using our `FileIO`.

### DIFF
--- a/docs/tutorials/duckdb.md
+++ b/docs/tutorials/duckdb.md
@@ -1,0 +1,102 @@
+# Querying benchmark results at scale with duckDB
+
+For a powerful way to query, filter, and visualize benchmark records, [duckdb](https://duckdb.org/) is a great choice.
+This page contains a quick tutorial for analyzing benchmark results with duckDB.
+
+## Prerequisites and installation
+
+To use duckdb, install the duckdb Python package by running `pip install --upgrade duckdb`.
+In this tutorial, we are going to be using the in-memory database only, but you can easily persist SQL views of records on disk as well.
+
+## Writing and ingesting benchmark records
+
+We consider the following easy benchmark example:
+
+```python
+--8<-- "examples/bq/benchmarks.py"
+```
+
+Running both of these benchmarks produces a benchmark record, which we can save to disk using the `FileIO` class.
+
+```python
+import nnbench
+from nnbench.context import GitEnvironmentInfo
+from nnbench.reporter.file import FileIO
+
+runner = nnbench.BenchmarkRunner()
+record = runner.run("benchmarks.py", params={"a": 1, "b": 1}, context=(GitEnvironmentInfo(),))
+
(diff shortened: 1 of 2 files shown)


nicholasjng:
I have not yet removed the `DuckDBReporter`, since it technically does more than what happens in this example, which is creating benchmark records from duckdb SQL queries.

I guess we are not in a hurry to remove the functionality though, so it can stay there without harming anyone. But as a reference on how to crunch/analyze records, this is an improvement over the duckDB reporter.

docs/tutorials/duckdb.md
```diff
@@ -0,0 +1,102 @@
...
+This page contains a quick tutorial for analyzing benchmark results with duckDB.
+
+## Prerequisites and installation
+
+To use duckdb, install the duckdb Python package by running `pip install --upgrade duckdb`.
+In this tutorial, we are going to be using the in-memory database only, but you can easily persist SQL views of records on disk as well.
+
+## Writing and ingesting benchmark records
+
+We consider the following easy benchmark example:
```
maxmynter, line 13:
```suggestion
We consider the following, easy benchmark example:
```

docs/tutorials/duckdb.md
```diff
@@ -0,0 +1,102 @@
...
+# │ varchar │ int64 │
+# ├─────────┼───────┤
+# │ prod    │     1 │
+# │ sum     │     2 │
+# └─────────┴───────┘
+```
+
+## Querying metadata directly in SQL by flattening the context struct
+
+By default, the benchmark context struct, which holds additional information about the benchmark runs, is inlined into the raw dictionary before saving it to a file.
```
maxmynter, line 58:
```suggestion
The benchmark context struct, holds additional information about the benchmark runs. By default, it is inlined into the raw dictionary before saving it to a file.
```

docs/tutorials/duckdb.md
```diff
@@ -0,0 +1,102 @@
...
+
+To improve, you can pass `ctxmode="flatten"` to the `FileIO.write()` method to flatten the context and inline all nested values instead.
+This comes at the expense of an inflated schema, i.e. more columns in the database.
+
+```python
+fio = FileIO()
+fio.write(record, "record.json", driver="ndjson", ctxmode="flatten")
+```
+
+In the example above, we used the `GitEnvironmentInfo` context provider to log some information on the git environment we ran our benchmarks in.
```
maxmynter, line 69: We should add some more info on context providers. Either here as a tangent, or in a separate guide. But I feel with providers, dicts, and Context objects the complexity warrants an explanation.

maxmynter (approved):
Nice :)

PR #92, opened 2024-02-26
Example Guide: Benchmark on Saved Models

### LINKED ISSUE #63: Add example: Implement benchmarking on saved artifacts
- [ ] Performance diffing by materializing older versions of a model and then comparing accuracy on a different evaluation set (needs design of artifact sourcing, download, materialization).

maxmynter:
This PR introduces artifact handling alongside a guide and example for these artifacts. The example concerns a bigger model than the other examples we have written so I believe it can serve as a nice case study as well. So if you see any clunky-ness w.r.t. to the nnbench usage, that can be a pointer to (a) improve the example, or (b) add nnbench functionality.

# Artifact Class
- Extend the `Artifact` class and move the loading logic to a `ArtifactLoader` to separate the concerns between loading and deserializing.
- Add an `ArtifactCollection` wrapper around an iterable to iterate through artifacts.

# Other nnbench changes
- `nnbench.runner` can take all contexts that the `Context.update` method can handle. This is useful if someone wants to ad-hoc add values in a dict or with a new `Context` to the context creation of a runner.
- (bugfix) Base reporter can now also handle when no custom formatters are supplied.

# Example
The example is a token classification task. It is mostly an adaptation from [here](https://huggingface.co/learn/nlp-course/chapter7/2?fw=pt). The logic is in the `examples/on_saved/src/training/training.py`. 

When people want to recreate the model training themselves they can copy it into a free GPU runtime on e.g. Google Colab, install the dependencies (`!pip install transformers[torch] datasets`) and run `!python training.py`. After About 5 minutes they have the necessary files ready for download. Alternatively they can get a model directly from Huggingface (e.g. [this](https://huggingface.co/dslim/bert-base-NER)).

Evaluation is executed by running `python runner.py <path-to-model-folder-1> <path-2> ...`.

The benchmarks are in the `benchmark.py`. Namely, accuracy, precision, recall, and f1 averaged and per label, plus some model metadata. We cache the evaluation processing where appropriate and use our `nnbench.parametrize` for label specific metrics

# Questions
- Especially around the dataset the `Artifact` wrapper feels a little artificial, especially w.r.t. the necessary preprocessing of the data to a dataloader. But maybe that is because Huggingface already provides a wrapper that is not available for custom datasets.
- I did not add a merging of Benchmark Records. I think that is more on the side of whatever backend we use. There I would want the run for specific models to be different benchmark records. An exception to this could be when i want deltas in the metrics between the current and a baseline model. In this case supplying the baseline to the reporter (and have it cached) feels like the most natural implementation. What do you think?

## Remaining ToDo's
- [x] Clean gitignore
- [x] Docstrings
- [x] Prepare Pyproject.toml
- [x] ~~Add common loaders (in a follow up PR?)~~ -> Follow up Issue (#97)
- [x] Add tokenizer saving to the `training.py` script
- [x] Write Guide
- [x] Clean History so we can merge with rebase (as this PR is quite big, I think that makes more sense than squashing).

Closes #63


src/nnbench/runner.py
```diff
@@ -187,7 +187,7 @@ def run(
         path_or_module: str | os.PathLike[str],
         params: dict[str, Any] | Parameters | None = None,
         tags: tuple[str, ...] = (),
-        context: Sequence[ContextProvider] = (),
+        context: Sequence[ContextProvider | dict[str, Any] | Context] = (),
```
nicholasjng, line 190: A sequence of contexts? I'm not sure that's correct. We take one `Context` object, and add/update it throughout our benchmarks.

Can you elaborate on why this was necessary?

examples/on_saved/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[build-system]
+requires = ["setuptools>=45", "setuptools-scm[toml]>=7.1"]
+build-backend = "setuptools.build_meta"
+
```
nicholasjng, line 4: I'm not sure about the practice to check in pyproject and requirements files for code examples anymore.

Could we instead make a note in the accompanying document that dependencies abc and xyz need to be installed? These files are largely useless for a reader of the docs that just follows along on his/her machine.

.gitignore
```diff
@@ -124,3 +124,6 @@ public/
 
 # NumPy zip archives created by the examples.
 *.npz
+
+# macOS
+.DS_Store
```
nicholasjng, line 129: I'm willing to check this in, but some people (i.e. those wanting to maintain a minimal project-wide .gitignore) will scream at you for this and tell you to put it into your machine's global .gitignore file.

Example: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
+# Benchmark on saved models
+There is a high likelihood that you, at some point, find yourself wanting to benchmark models that were trained previously.
+In this guide we will walk through how we can accomplish this with nnbench.
+We will start with an aside that talks through the setup of the example we will use in this guide.
+If you are only interested in the application of nnbench, you can skip this section.
```
nicholasjng, line 5: I would put these into the NER section below, and leave out the exposition (L4) or maybe reduce it to "This is a very short primer on the experiment setup".

examples/on_saved/src/benchmark.py
```diff
@@ -0,0 +1,223 @@
+import tempfile
```
nicholasjng, line 1: Can you rename that directory to something more expressive, like "examples/artifact-benchmarking"?

examples/on_saved/src/runner.py
```diff
@@ -0,0 +1,96 @@
...
+            remove_columns=dataset.column_names,
+        )
+        dataloader = DataLoader(
+            tokenized_dataset,
+            shuffle=False,
+            collate_fn=DataCollatorForTokenClassification(tokenizer, padding=True),
+            batch_size=8,
+        )
+        result = runner.run(
+            "benchmark.py",
```
nicholasjng, line 75: I would be interested in what happens here - does the runner collect and clear everytime or does the collection happen once and is then reused in the process?

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Example: Named Entity Recognition
+There are several reasons why you want to retrieve saved models for benchmarking.
```
nicholasjng, line 8:
```suggestion
There are lots reasons why you could want to retrieve saved models for benchmarking. 
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
+# Benchmark on saved models
+There is a high likelihood that you, at some point, find yourself wanting to benchmark models that were trained previously.
+In this guide we will walk through how we can accomplish this with nnbench.
+We will start with an aside that talks through the setup of the example we will use in this guide.
+If you are only interested in the application of nnbench, you can skip this section.
+
+## Example: Named Entity Recognition
+There are several reasons why you want to retrieve saved models for benchmarking. 
+Among others, these are reviewing the work of colleagues, comparing experimental performances to a benchmark, or dealing with models that require significant compute such that in place retraining is impractical.
```
nicholasjng, line 9:
```suggestion
Among them are reviewing the work of colleagues, comparing experimental performances to an existing benchmark, or dealing with models that require significant compute such that in-place retraining is impractical.
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+For this example, we deal with a named entity recognition (NER) model that is based on the pre-trained encoder-decoder transformer [BERT](https://arxiv.org/abs/1810.04805).
+The model is trained on the [CoNLLpp dataset](https://huggingface.co/datasets/conllpp) which consists of sentences from news stories where words were tagged with Person, Organization, Location, or Miscellaneous if they referred to entities. 
+Words were assigned an out of entity label if they did not represented an entity.
```
nicholasjng, line 12:
```suggestion
Words were assigned an out-of-entity label if they do not represent an entity.
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Model Training
+You find the code to train the model in the nnbench [repository](https://github.com/aai-institute/nnbench) in the directory `examples/on_saved/src/training/training.py`.
+If you want to skip running the training script but still want to reproduce this example, you can take any BERT model fine tuned for NER with the CoNLL- dataset family.
```
nicholasjng, line 16:
```suggestion
If you want to skip running the training script but still want to reproduce this example, you can take any BERT model fine tuned for NER with the CoNLL dataset family.
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+You find many on the Huggingface model hub, for example [this one](https://huggingface.co/dslim/bert-base-NER). You need to download the `model.safetensors`, `config.json`, `tokenizer_config.json`, and `tokenizer.json`.
```
nicholasjng, line 17:
```suggestion
You find many on the Huggingface model hub, for example [this one](https://huggingface.co/dslim/bert-base-NER). You need to download the `model.safetensors`, `config.json`, `tokenizer_config.json`, and `tokenizer.json` files.
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+If you want to train your own model, continue below. 
+
+
```
nicholasjng, line 20:
```suggestion
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+Hence, if you are working with a different GPU instance, make sure to install everything from the `pyproject.toml` in the `examples/on_saved` folder. 
+Next you can execute the `training.py` with `!python training.py`.
+This will train two BERT models ("bert-base-uncased" and "distilbert-base-uncased") which we can compare using nnbench. 
+If you want, you can adapt the training script to train other models by editing the tuples in the `tokenizers_and_models` list at the bottom of the training script. 
+The training of the models takes around 10 minutes.
+Once it is done, download the respective files and save them to your disk.
+They should be the same mentioned above. 
+We will need the path to the files for benchmarking later.
+
+
```
nicholasjng, line 38:
```suggestion
```

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+## Example: Named Entity Recognition
+There are several reasons why you want to retrieve saved models for benchmarking. 
+Among others, these are reviewing the work of colleagues, comparing experimental performances to a benchmark, or dealing with models that require significant compute such that in place retraining is impractical.
+For this example, we deal with a named entity recognition (NER) model that is based on the pre-trained encoder-decoder transformer [BERT](https://arxiv.org/abs/1810.04805).
+The model is trained on the [CoNLLpp dataset](https://huggingface.co/datasets/conllpp) which consists of sentences from news stories where words were tagged with Person, Organization, Location, or Miscellaneous if they referred to entities. 
+Words were assigned an out of entity label if they did not represented an entity.
+
+### Model Training
+You find the code to train the model in the nnbench [repository](https://github.com/aai-institute/nnbench) in the directory `examples/on_saved/src/training/training.py`.
```
nicholasjng, line 15: Let's adjust the folder name here and subsequently, too, if we decide to do so.

docs/tutorials/on_saved.md
```diff
@@ -0,0 +1,116 @@
...
+They should be the same mentioned above. 
+We will need the path to the files for benchmarking later.
+
+
+### The Benchmarks
+The benchmarking code is found in the `examples/on_saved/benchmark.py`.
+We calculate precision, recall, accuracy, and f1 scores for the whole test set and specific labels.
+Additionally, we obtain information about the model such as its memory footprint and inference time.
+We are not walking through the whole file but instead point out certain design choices as an inspiration to you. 
+If you are interested in a more detailed walkthrough of the setup of benchmarks, you find it [here](../guides/benchmarks.md).
```
nicholasjng, line 44:
```suggestion
If you are interested in a more detailed walkthrough on how to set up benchmarks, you can find it [here](../guides/benchmarks.md).
```

nicholasjng (commented):
Additional comments:

- The lazy artifact loading is really sweet!
- We could define a `functools.partial` alias for the parametrization decorator with all the class labels to refactor the raw injections a little bit.
- Could you share a screenshot on how the results look in a table?

src/nnbench/types.py
```diff
@@ -93,42 +96,148 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+        ------
+        ValueError
+            If the artifact has not been deserialized yet.
+        """
         if self._value is None:
             raise ValueError(
                 f"artifact has not been instantiated yet, "
-                f"perhaps you forgot to call {self.__class__.__name__}.materialize()?"
+                f"perhaps you forgot to call {self.__class__.__name__}.deserialize()?"
             )
```
AdrianoKF, line 188: Should this automatically materialize the artifact instead? From a user's point of view, this would make the deserialization completely transparent.

src/nnbench/runner.py
maxmynter: I wanted to pass information to the context parameter of the runner. Specifically the name of the validation data set that I used (there are multiple versions of the CoNLL dataset).

As the `Context.update` also handles `ContextProvider | dict[str, Any] | "Context"`, I thought it makes sense that the runner can handle all these types as well since the runner uses `Context.update()` to handle the provided context.

examples/on_saved/pyproject.toml
maxmynter: @janwillemkl remarked on the Streamlit integration to prepare a pyproject file for the example. That is why I prepared one here. 

It is essentially a `pip install datasets transformers[torch] nnbench`, so it would also work without a requirements file, I guess.

.gitignore
maxmynter: I can do that instead and remove the `.DS_store` here. I did not know about global `.gitignore` before.

src/nnbench/types.py
maxmynter: I think that makes sense. 
Automatic serialization could also simplify the logic in the `ArtifactCollection` class as its purpose - aside from being an iteration wrapper - is to make sure artifacts are deserialized when we want to use them. 

As you wrote the initial logic, what do you think @nicholasjng ?

maxmynter:
> We could define a functools.partial alias for the parametrization decorator with all the class labels to refactor the raw injections a little bit.

Agree

> Could you share a screenshot on how the results look in a table?

There you go :) 

Screenshot 2024-03-06 at 12 37 01

src/nnbench/runner.py
nicholasjng: Good thinking. I'm not a fan of this typing, which is because the `Context.update` typing is too permissive.

Since the context is supposed to group different value sets on top-level keys (e.g. git info under "git"), I think the following should be true:

- `Context.update` merges two `Context` objects, so "other" needs to be a `Context`.
- `Context.add(provider)` inserts a context provider's result into the context under a single top-level context key.
- We drop raw dict support entirely, outside of `Context.make()` and the constructor.

Then, the typing of `context` in `BenchmarkRunner.run()` becomes `Context | Sequence[ContextProvider]` - the former if you want to supply your own hand-built context, the latter if you want it assembled from atomic providers. How does that sound?

src/nnbench/types.py
nicholasjng: All good with me.

maxmynter:
Using `functools.partial` does not work straightforward. The collection then does not recognize the parametrized benchmarks anymore. Therefore, this is out of scope for this PR.

examples/on_saved/src/runner.py
maxmynter: Collection happens only in the first run :)

src/nnbench/runner.py
```diff
@@ -187,7 +187,7 @@ def run(
         path_or_module: str | os.PathLike[str],
         params: dict[str, Any] | Parameters | None = None,
         tags: tuple[str, ...] = (),
-        context: Sequence[ContextProvider] = (),
+        context: Sequence[ContextProvider | Context] = (),
```
nicholasjng, line 190: No, I meant `Context | Sequence[ContextProvider]`.

If it's the former, you say `ctx = context` (in L239), if it's the latter, you call `ctx.add(provider)` in the loop over the providers.

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+
+    @property
     def value(self) -> T:
+        """
+        Returns the deserialized artifact value.
+
+        Returns
+        -------
+        T
+            The deserialized value of the artifact.x
```
nicholasjng, line 177:
```suggestion
            The deserialized value of the artifact.
```

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+            self._artifacts.append(artifacts)
+
+    def __iter__(self) -> Iterator[Artifact[T]]:
+        """
+        Iterator over deserialized Artifacts in the collection.
+
+        Yields
+        -------
+        Iterator[Artifact[T]]
+            An iterator over deserialized Artifacts in the collection.
```
nicholasjng, line 213:
```suggestion
            An iterator over deserialized artifacts in the collection.
```

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+            The Artifact instance(s) to add to the collection.
+        """
+        if isinstance(artifacts, Iterable):
+            self._artifacts.extend(artifacts)
+        else:
+            self._artifacts.append(artifacts)
+
+    def __iter__(self) -> Iterator[Artifact[T]]:
+        """
+        Iterator over deserialized Artifacts in the collection.
```
nicholasjng, line 208:
```suggestion
        Iterator over deserialized artifacts in the collection.
```

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+        -------
+        Iterator[Artifact[T]]
+            An iterator over deserialized Artifacts in the collection.
+        """
+        for artifact in self._artifacts:
+            yield artifact
+
+    def values(self) -> Iterator[T]:
+        """
+        Iterator over the values of deserialized Artifacts in the collection.
```
nicholasjng, line 220:
```suggestion
        Iterator over the values of deserialized artifacts in the collection.
```

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
+
+        Yields
+        ------
+        Iterator[T]
+            An iterator over the values of the deserialized Artifacts.
```
nicholasjng, line 225:
```suggestion
            An iterator over the values of the deserialized artifacts.
```

src/nnbench/types.py
```diff
@@ -93,42 +96,138 @@ def expand(cls, bms: list[dict[str, Any]]) -> BenchmarkRecord:
...
-    @classmethod
-    def materialize(cls) -> "Artifact":
-        """Load the artifact from storage."""
-        raise NotImplementedError
+    @abstractmethod
+    def deserialize(self) -> None:
+        """Deserialize the artifact."""
+
+    def is_deserialized(self) -> bool:
+        """ "Checks if the Artifact is already deserialized."""
```
nicholasjng, line 160:
```suggestion
        """ "Checks if the artifact is already deserialized."""
```

src/nnbench/context.py
```diff
@@ -276,21 +276,27 @@ def items(self, sep: str = ".") -> Iterator[tuple[str, Any]]:
         """
         yield from self._ctx_items(d=self._data, prefix="", sep=sep)
 
-    def update(self, other: ContextProvider | dict[str, Any] | "Context") -> None:
+    def add(self, provider: ContextProvider) -> None:
+        """
+        Adds data from a provider to the Context.
```
nicholasjng, line 281:
```suggestion
        Adds data from a provider to the context.
```

src/nnbench/context.py
```diff
@@ -276,21 +276,27 @@ def items(self, sep: str = ".") -> Iterator[tuple[str, Any]]:
...
+
+        Parameters
+        ----------
+        provider: ContextProvider
+            The context provider; a callable that returns a dict with context values.
```
nicholasjng, line 286:
```suggestion
            The provider to inject into this context.
```

src/nnbench/context.py
```diff
@@ -276,21 +276,27 @@ def items(self, sep: str = ".") -> Iterator[tuple[str, Any]]:
...
         """
         Updates the context.
-        If `sep` is None, perform shallow update, deep update otherwise.
 
         Parameters
         ----------
-        other : ContextProvider | dict[str, Any] | "Context"
-            The data to update the context with. This can be a dictionary, a Context instance, or a provider.
+        other :  "Context"
+            The other Context to update this Context with.
```
nicholasjng, line 297:
```suggestion
            The other context to update this context with.
```

nicholasjng (commented):
In general:

If you use the english word of a type (say, Context), please use lower-case spelling, since the word and not the type abstraction is meant.

Example: The context to update -> context is lowercase, since it means the english word "context", represented by the `Context` object.

src/nnbench/runner.py
maxmynter: In this implementation (using `add` we then do not throw any error on multiple values for context key since the `add` (and `update`) methods use `dict.update`, which overwrites keys, under the hood.

Are we fine with that?

src/nnbench/runner.py
```diff
@@ -236,16 +236,20 @@ def run(
...
-            if duplicates:
-                dupe, *_ = duplicates
-                raise ValueError(f"got multiple values for context key {dupe!r}")
-            ctx.update(ctxval)
+                # we do not allow multiple values for a context key.
+                duplicates = set(ctx.keys()) & set(ctx_cand.keys())
+                if duplicates:
+                    dupe, *_ = duplicates
+                    raise ValueError(f"got multiple values for context key {dupe!r}")
+                ctx.update(ctx_cand)
```
nicholasjng, line 252: This method is suboptimal with the unique key requirement, but we could implement a key check under the hood in `Context.update()`, and add raw providers here in the loop with `Context.add()`. That's at most a follow-up, though.

nicholasjng (approved):

src/nnbench/runner.py
maxmynter: Nice, I can open an issue for that.

PR #98, opened 2024-03-07
Add a lightning demo to the README.md

nicholasjng:
Our package is structured well enough that this example is small, so it does not take lots of processing new information to understand.

---------------

For reviewers:

* Does this appropriately convey what we're doing?
* After reading this, could you write the code to add and run a `sum` benchmark without changing the `runner = nnbench.BenchmarkRunner()` and subsequent lines?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,40 @@ pip install nnbench
 poetry add nnbench
 ```
 
+## A ⚡️- quick demo
+
+To understand how nnbench works, you can run the following in your Python interpreter:
+
+```python
+# example.py
+import nnbench
+
+
+@nnbench.benchmark
+def product(a: int, b: int) -> int:
+    return a * b
+
+
+@nnbench.benchmark
+def power(a: int, b: int) -> int:
+    return a ** b
+
+
+runner = nnbench.BenchmarkRunner()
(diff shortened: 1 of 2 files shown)


nicholasjng:
BTW this needs a change in the benchmark reporter's custom formatters, which is contained in #92.

maxmynter (commented):
From this example I understand how to set up a benchmark. I don't know if I would necessarily know why I would want to use it. 

Should we add a sentence at the end along the lines of: 

"To move beyond toy examples, you can pass your model object as a parameter to benchmarks which calculate the metrics you are interested in. For more advanced usages [...]"

nicholasjng:
Good idea - however, the setup with an actual model will likely make the example 4-5x larger, which is not what we want for a lightning demo.

I agree with the note for real-world uses, though. Would you be happy with a link to a doc where we use models? (Suggestions welcome)

nicholasjng:
I added links to the quickstart and the examples, and added anchor links on the examples indexpage in https://github.com/aai-institute/nnbench/pull/98/commits/0ac0592b41d65e02489501385e7b51297443eb6b.

maxmynter (approved):
Nice one :)

PR #100, opened 2024-03-07
Implement a FSspec based ArtifactLoader

### LINKED ISSUE #94: Implement fsspec artifact loader for nnbench
- [ ] Functionality for loading models from any fsspec supported url has been implemented.
- [ ] Set fsspec loader as a default.

maxmynter:
Closes #94 

# Questions
- Should `fsspec` become a first class dependecy such as tabulate? Or do we import it conditionally in the `FilepathArtifactLoader`?

### DIFF
--- a/.github/workflows/lint-and-test.yaml
+++ b/.github/workflows/lint-and-test.yaml
@@ -58,7 +58,6 @@ jobs:
     - name: Run tests on oldest supported Python
       run: |
         python -m pip install ".[dev]"
-        pip install fsspec
         pytest
   docs:
     name: Publish latest documentation for nnbench
--- a/.github/workflows/lint-and-test.yaml
+++ b/.github/workflows/lint-and-test.yaml
@@ -58,6 +58,7 @@ jobs:
     - name: Run tests on oldest supported Python
       run: |
         python -m pip install ".[dev]"
+        pip install fsspec
         pytest
   docs:
     name: Publish latest documentation for nnbench
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ dev = [
     "psutil",
     "pyarrow",
     "pytest>=7.4.0",
+    "fsspec"
 ]
 docs = [
     "black",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ strict_optional = false
 warn_unreachable = true
 
 [[tool.mypy.overrides]]
-module = ["tabulate", "yaml"]
+module = ["tabulate", "yaml", "fsspec"]
 ignore_missing_imports = true
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["tabulate", "fsspec"]
+dependencies = ["tabulate"]
 
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["tabulate"]
+dependencies = ["tabulate", "fsspec"]
 
 dynamic = ["version"]
 
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -12,12 +12,24 @@ from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass, field
 from pathlib import Path
 from tempfile import mkdtemp
-from typing import Any, Callable, Generic, Iterable, Iterator, Literal, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Generic,
+    Iterable,
+    Iterator,
+    Literal,
+    TypeVar,
+)
 
 from nnbench.context import Context
 
 T = TypeVar("T")
 Variable = tuple[str, type, Any]
+
+if TYPE_CHECKING:
+    import fsspec
 
 
 def NoOp(**kwargs: Any) -> None:
@@ -142,11 +154,6 @@ class FilePathArtifactLoader(ArtifactLoader):
         The local directory to which remote artifacts will be downloaded. If provided, the model data will be persisted. Otherwise, local artifacts are cleaned.
     storage_options : dict[str, Any] | None
         Storage options for remote storage.
-
-    Raises
-    ------
-    ImportError
-        When the fsspec module is not installed.
     """
 
     def __init__(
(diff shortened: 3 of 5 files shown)


src/nnbench/types.py
```diff
@@ -110,22 +114,52 @@ class LocalArtifactLoader(ArtifactLoader):
...
+    def load(self) -> Path:
+        """
+        Loads the artifact, downloading it if necessary, and returns the local path.
+
+        Returns
+        -------
+        Path
+            The path to the artifact on the local filesystem.
+        """
+        fs = spec.AbstractFileSystem()
```
nicholasjng, line 160: The filesystem should be constructed directly from the URI, see e.g. https://github.com/nicholasjng/shelf/blob/4a628c17e678524a49cb327238c3fb7924e0cb9d/src/shelf/util.py#L12.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["tabulate"]
+dependencies = ["tabulate", "fsspec"]
```
nicholasjng, line 34: In my opinion, we should make this optional (i.e. by importing `fsspec` only locally in the artifact load method), and facilitate local loading with builtin methods.

src/nnbench/types.py
```diff
@@ -6,10 +6,17 @@
...
 import os
+import shutil
+import weakref
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass, field
+from pathlib import Path
+from tempfile import mkdtemp
 from typing import Any, Callable, Generic, Iterable, Iterator, Literal, TypeVar
 
+from fsspec import core, filesystem, utils
```
nicholasjng, line 17: Please import the specific APIs from these submodules, not the whole modules themselves.

tests/test_artifacts.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
+
+from nnbench.types import FilePathArtifactLoader
+
+
+def test_load_local_file(local_file: Path, tmp_path: Path) -> None:
+    test_dir = tmp_path / "test_load_dir"
+    loader = FilePathArtifactLoader(str(local_file), str(test_dir))
```
nicholasjng, line 8: This should work without the string casts.

src/nnbench/types.py
```diff
@@ -110,22 +117,72 @@ class LocalArtifactLoader(ArtifactLoader):
...
+    to_localdir : str | os.PathLike[str] | None
+        The local directory to which remote artifacts will be downloaded. If provided, the model data will be persisted. Otherwise, local artifacts are cleaned.
+    storage_options : dict[str, Any] | None
+        Storage options for remote storage.
+    """
+
+    def __init__(
+        self,
+        path: str | os.PathLike[str],
+        to_localdir: str | os.PathLike[str] | None = None,
```
nicholasjng, line 153: This should have a different name, it currently sounds like a bool. Maybe `destination` / `basedir` ?

src/nnbench/types.py
```diff
@@ -110,22 +117,72 @@ class LocalArtifactLoader(ArtifactLoader):
...
+        Loads the artifact, downloading it if necessary, and returns the local path.
+
+        Returns
+        -------
+        Path
+            The path to the artifact on the local filesystem.
+        """
+        fs, _, _ = core.get_fs_token_paths(self.source_path)
+        if isinstance(fs, LocalFileSystem):
+            return Path(self.source_path).resolve()
```
nicholasjng, line 177: Is this caching a la "check if it exists, only download if it doesn't"? (Otherwise it's probably currently missing)

nicholasjng (commented):
What is the story after I downloaded the remote files to my machine? I mean, what do I have to use to instantiate my model from the files?

pyproject.toml
maxmynter: Agree

src/nnbench/types.py
maxmynter: More or less.

My understanding is, that this class, as a general filesystem class, should also be able to use local filepaths.
In that case, we don't want to download the data.

If it is a remote path, we download and put it either (a) a specified local dir,  or (b) a tmp dir. 

If a) the files are persisted on disk and the filepath can be used (by the user) in other places. 

If b) the cleanup task deletes the tmp dir and the model to not clutter the filesystem. 

That is how it is currently implemented. Do you agree with this?

.github/workflows/lint-and-test.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
     - name: Run tests on oldest supported Python
       run: |
         python -m pip install ".[dev]"
+        pip install fsspec
```
nicholasjng, line 61: No, just add it to the dev extra.

src/nnbench/types.py
```diff
@@ -110,22 +114,98 @@ class LocalArtifactLoader(ArtifactLoader):
...
+
+            self._LocalFileSystem = LocalFileSystem
+
+            from fsspec.utils import get_protocol
+
+            self._get_protocol = get_protocol
+
+            from fsspec import filesystem
+
+            self._filesystem = filesystem
```
nicholasjng, line 173: These methods will not show up in language server completions / linters. Instead, you should import them in the class API before they are actually used. Example:

```python
class A:
    def proto(url: str) -> str:
        from fsspec.utils import get_protocol
        return get_protocol(url)
```

src/nnbench/types.py
```diff
@@ -110,22 +114,98 @@ class LocalArtifactLoader(ArtifactLoader):
...
+        except ImportError:
+            raise ImportError(
+                "The FilePathArtifactLoader depends on fsspec. Do you have it installed?"
+            )
```
nicholasjng, line 177: That's not helpful, you should instead give an install instruction:

```suggestion
            raise ImportError(
                "class {self.__class__.__name__} requires `fsspec` to be installed. You can install it by running `python -m pip install --upgrade fsspec`"
            )
```

src/nnbench/types.py
```diff
@@ -110,22 +114,98 @@ class LocalArtifactLoader(ArtifactLoader):
...
+    def load(self) -> Path:
+        """
+        Loads the artifact, downloading it if necessary, and returns the local path.
+
+        Returns
+        -------
+        Path
+            The path to the artifact on the local filesystem.
+        """
+        fs, _, _ = self._get_fs_token_paths(self.source_path)
```
nicholasjng, line 198: Since you're ignoring the other two outputs, you could also use `fs = fsspec.filesystem(self.source_path, **self.storage_options)` (or with the options in another set of parentheses, not sure about the calling convention).

src/nnbench/types.py
```diff
@@ -110,22 +114,98 @@ class LocalArtifactLoader(ArtifactLoader):
...
+        if isinstance(fs, self._LocalFileSystem):
+            return Path(self.source_path).resolve()
```
nicholasjng, line 200: You might want to copy files into your target folder even if they are local? Seems like you could just omit the branch here.

src/nnbench/types.py
```diff
@@ -110,22 +126,90 @@ class LocalArtifactLoader(ArtifactLoader):
...
+    def _get_protocol(self, path: str) -> str:
+        """Lazy loading of fsspec.utils.get_protocol"""
+        try:
+            from fsspec.utils import get_protocol
+
+            return get_protocol(path)
+        except ImportError:
+            raise ImportError(
+                "class {self.__class__.__name__} requires `fsspec` to be installed. You can install it by running `python -m pip install --upgrade fsspec`"
+            )
```
nicholasjng, line 184: Two things here:

1) The single use in `load()` below does not warrant extracting these into their own APIs - please just import the fsspec versions both inline in `load()`, and scrap the private class methods.
2) You can factor out the import error with a simple import guard at the top of the file, e.g. like so:

```python
# this towards the top, after the imports.

try:
    import fsspec
    HAS_FSSPEC = True
except ImportError:
    HAS_FSSPEC = False

...
# later, in FilePathArtifactLoader.__init__():
...
if not HAS_FSSPEC:
    raise ModuleNotFoundError("class blah requires `fsspec`...")
```

.github/workflows/lint-and-test.yaml
nicholasjng: Please remove this line, then.

src/nnbench/types.py
```diff
@@ -6,15 +6,31 @@
...
+    TypeVar,
+)
 
 from nnbench.context import Context
 
 T = TypeVar("T")
 Variable = tuple[str, type, Any]
 
+if TYPE_CHECKING:
+    import fsspec
```
nicholasjng, line 32: Is this mypy complaining? You could also add `fsspec` to the list of modules for which to ignore missing imports (in the pyproject.toml) if you like, since most of its interfaces are not great typing-wise.

nicholasjng (approved):

PR #102, opened 2024-03-08
Hotfix: Adjust snippet paths

nicholasjng:
These misconfigured paths were the reason why the code snippets were not shown.


nicholasjng:
This doc deserves some more polishing, especially in the wall of text around the training.

PR #103, opened 2024-03-08
Add Parameters to Benchmarks

### LINKED ISSUE #101: Benchmark results do not contain parameters
Reported by @samuelburbulla (thanks!).

Simply put: In the lightning demo,

```python
import nnbench


@nnbench.benchmark
def product(a: int, b: int) -> int:
    return a * b


@nnbench.benchmark
def power(a: int, b: int) -> int:
    return a ** b


runner = nnbench.BenchmarkRunner()
# run the above benchmarks with the parameters `a=2, b=10`...
record = runner.run("__main__", params={"a": 2, "b": 10})
rep = nnbench.BenchmarkReporter()
rep.display(record)  # ...and print the results to the terminal.

# results in a table like the following:
# name     function    date                   value    time_ns
# -------  ----------  -------------------  -------  ---------
# product  product     2024-03-07T10:14:21       20       1000
# power    power       2024-03-07T10:14:21     1024        750
```

we would expect to find the actual parameter values (`a=2, b=10`) in the table, since they are important to contextualize results, and provide filtering opportunities (for example, "show me each `product` result where `a=2`").

As before, we should think about how to deal with these parameters. Inserting them raw into the result record is the simplest solution, but that again balloons up the schema of multiple concurring experiments taking different sets of parameters.

In contrast, inserting the parameters as a struct/dictionary makes access a little more difficult, but keeps the schema easier (and some modern SQL engines support dotted access for struct columns, see e.g. the duckDB example).

Checklist:
- [ ] Provide parameter struct as a key in the benchmark record, inserted in `runner.run()` (take the union of `params` and the defaults from the benchmark interface).
- [ ] Show an application (maybe as a screenshot of application code or a unit test) of filtering a record by parameter values.

maxmynter:
Closes #101 

Add the Parameters to a 'BenchmarkRecord' by appending it to the results. Omit the type information from the `Variable` object and only retain variable name and value.

### DIFF
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -264,7 +264,7 @@ class BenchmarkRunner:
                 "date": datetime.now().isoformat(timespec="seconds"),
                 "error_occurred": False,
                 "error_message": "",
-                "parameters": {**bmdefaults, **bmparams},
+                "parameters": bmdefaults | bmparams,
             }
             try:
                 benchmark.setUp(**bmparams)
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -89,7 +89,9 @@ class BenchmarkRunner:
                     )
 
                 if typ == empty:
-                    logger.debug(f"parameter {name!r} untyped in benchmark {bm.fn.__name__}().")
+                    logger.debug(
+                        f"parameter {name!r} untyped in benchmark {bm.fn.__name__}()."
+                    )
 
                 if name in allvars:
                     currvar = allvars[name]
@@ -134,7 +136,9 @@ class BenchmarkRunner:
         """Clear all registered benchmarks."""
         self.benchmarks.clear()
 
-    def collect(self, path_or_module: str | os.PathLike[str], tags: tuple[str, ...] = ()) -> None:
+    def collect(
+        self, path_or_module: str | os.PathLike[str], tags: tuple[str, ...] = ()
+    ) -> None:
         # TODO: functools.cache this guy
         """
         Discover benchmarks in a module and memoize them for later use.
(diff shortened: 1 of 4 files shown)


src/nnbench/runner.py
```diff
@@ -255,6 +255,7 @@ def run(
         results: list[dict[str, Any]] = []
         for benchmark in self.benchmarks:
             bmparams = {k: v for k, v in dparams.items() if k in benchmark.interface.names}
+            bmdefaults = {k: v for (k, t, v) in benchmark.interface.variables}
```
nicholasjng, line 258:
```suggestion
            bmdefaults = {k: v for (k, _, v) in benchmark.interface.variables}
```

src/nnbench/runner.py
```diff
@@ -263,6 +264,7 @@ def run(
                 "date": datetime.now().isoformat(timespec="seconds"),
                 "error_occurred": False,
                 "error_message": "",
+                "parameters": {**bmdefaults, **bmparams},
```
nicholasjng, line 267: Does this ensure that the defaults are overwritten by the `bmparams`?

I think something like `bmdefaults.update(bmparams)` would be easier to read.

nicholasjng (commented):
Great job!

src/nnbench/runner.py
nicholasjng: I think this does what it needs to do: For any variable, if the default is used, it's not in `bmparams`, so the update does nothing - and for every overridden default, the `update()` overwrites with the argument given in the params dict.

So the order should be defaults, then update with params - which is what I suggested (you can also write `bmdefaults | bmparams`).

nicholasjng: If you need a mental model, it helps to think about what happens once the `params` are passed into the function - its default keyword arguments are already there, and everything else gets overridden by the incoming parameters. So it is indeed `bmdefault.update(bmparams)` we're looking for here (you can even use that as the dict value, since we do not explicitly use it again afterwards).

PR #109, opened 2024-03-12
Remove the `ArtifactCollection` class

### LINKED ISSUE #108: Remove `ArtifactCollection` Wrapper
As we added in-place deserialization in the `Artifact` class, the `ArtifactCollection` is a wrapper around a list with no added functionality. 
We should therefore remove it from 
- [x] types.py
- [x] guide
- [x] example

maxmynter:
It is redundant as we have added the deserialization logic in the `Artifact.value` call.
It was therefore only a somewhat useless wrapper around a list. 
Removed mentions from docs as well.

Closes #108

### DIFF
--- a/examples/artifact_benchmarking/src/runner.py
+++ b/examples/artifact_benchmarking/src/runner.py
@@ -53,12 +53,10 @@ def main(model_paths: list[str]) -> None:
         TokenClassificationModel(loader=types.LocalArtifactLoader(path)) for path in model_paths
     ]
 
-    model_store = types.ArtifactCollection(*models)
-
     conllpp = ConllValidationData(ConllppLoader("conllpp", split="validation"))
     conllpp.deserialize()
 
-    for model, tokenizer in model_store.values():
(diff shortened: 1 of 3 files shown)


examples/artifact_benchmarking/src/runner.py
```diff
@@ -53,12 +53,10 @@ def main(model_paths: list[str]) -> None:
         TokenClassificationModel(loader=types.LocalArtifactLoader(path)) for path in model_paths
     ]
 
-    model_store = types.ArtifactCollection(*models)
-
     conllpp = ConllValidationData(ConllppLoader("conllpp", split="validation"))
     conllpp.deserialize()
 
-    for model, tokenizer in model_store.values():
+    for model, tokenizer in [model.value for model in models]:
```
nicholasjng, line 59:
```suggestion
    for mod in models:
        model, tokenizer = mod.value
```

nicholasjng (commented):
Nice!

PR #110, opened 2024-03-13
Add checksum verification to artifacts and loaders

nicholasjng:
In preparation for the string representation of artifacts that can be saved to other places while keeping reproducibility.

Missing pieces:

- [ ] Tests
- [ ] `Artifact.to_json()/to_dict()` hook in preparation for transforms.

### DIFF
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -111,14 +111,10 @@ class BenchmarkRecord:
 
 class ArtifactLoader:
     @abstractmethod
-    def load(
-        self, rpath: str | os.PathLike[str], checksum: str | None = None
-    ) -> os.PathLike[str]:
+    def load(self, rpath: str | os.PathLike[str], checksum: str | None = None) -> os.PathLike[str]:
         """Load the artifact."""
 
-    def verify(
-        self, rpath: str | os.PathLike[str], expected: str, blocksize: int = 2**22
-    ) -> None:
+    def verify(self, rpath: str | os.PathLike[str], expected: str, blocksize: int = 2**22) -> None:
         """Verify an artifact path checksum against an expected value."""
 
 
@@ -150,9 +146,7 @@ class PathArtifactLoader(ArtifactLoader):
             from fsspec import AbstractFileSystem, filesystem
             from fsspec.utils import get_protocol
 
-            fs: AbstractFileSystem = filesystem(
-                get_protocol(str(rpath)), **self.storage_options
-            )
+            fs: AbstractFileSystem = filesystem(get_protocol(str(rpath)), **self.storage_options)
 
             if not fs.exists(rpath):
                 raise FileNotFoundError(rpath)
@@ -178,9 +172,7 @@ class PathArtifactLoader(ArtifactLoader):
                 self.verify(rpath, checksum)
             return Path(rpath).resolve()
 
-    def verify(
-        self, rpath: str | os.PathLike[str], expected: str, blocksize: int = 2**22
-    ) -> None:
+    def verify(self, rpath: str | os.PathLike[str], expected: str, blocksize: int = 2**22) -> None:
         with open(rpath, "rb") as f:
             chunk = f.read(blocksize)
             while chunk:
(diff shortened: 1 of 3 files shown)


nicholasjng:
Question (cc @maxmynter): Why is the artifact loader tied to a specific path? It seems fairly restrictive, without any way to switch out the path , and `load()` not taking the path as an argument, either.

Are there any design hurdles to making `path` (and now optionally `checksum`) part of the `Loader.load()` API?

Bonus: Does it make sense to pass a loader to the `runner.run()` (or the BenchmarkRunner itself) that is given to each artifact passed in `params` by default?

src/nnbench/types.py
```diff
@@ -130,12 +126,49 @@ class LocalArtifactLoader(ArtifactLoader):
     """
 
     def __init__(self, path: str | os.PathLike[str]) -> None:
-        self._path = path
+        import hashlib
+
+        self._path = Path(path)
+        self._hash = hashlib.md5(usedforsecurity=False)
+
+    def verify(self, expected: str, blocksize: int = 2**22) -> None:
```
maxmynter, line 134: Should we add the usage of `verify` to the guide (`/docs/guides/artifact.md`)?

I can also take care of that.

src/nnbench/types.py
```diff
@@ -149,11 +182,18 @@ class FilePathArtifactLoader(ArtifactLoader):
     Parameters
     ----------
     path : str | os.PathLike[str]
-        The path to the artifact, which can include a protocol specifier (like 's3://') for remote access.
```
maxmynter, line 152: How many columns does a docstring line have at maximum?
Or do you just linebreak after every sentence?

src/nnbench/types.py
```diff
@@ -162,6 +202,13 @@ def __init__(
         destination: str | os.PathLike[str] | None = None,
         storage_options: dict[str, Any] | None = None,
     ) -> None:
+        import importlib.util
+
+        if importlib.util.find_spec("fsspec") is None:  #
+            raise ModuleNotFoundError(
+                f"class {self.__class__.__name__}() requires `fsspec` to be installed. "
+                "You can install it by running `python -m pip install --upgrade fsspec`"
+            )
```
maxmynter, line 211: Cool. That is way nicer than the top file import guard.

maxmynter (commented):
## Initial rationale for `Loader`

We thought about the `ArtifactLoader` instance to be tied to a specific artifact. That way you would create another instance when you want a different one. 

E.g. you have an `mlflow` model registry (pseudocode).

```python 
import mlflow
from nnbench.types import ArtifactLoader
from tempfile import mkdtemp

class MlflowRegistryLoader(ArtifactLoader):
    def __init__(self, tracking_uri, model_name, model_version):
        self.tracking_uri = tracking_uri
        self.model_name = model_name
        self.model_version = model_version

    def load(self):
        mlflow.set_trackin_uri(self.tracking_uri)
        model_uri = mlflow.get_model_version_uri(self.model_name, self.model_version)
        local_dir = mkdtemp()
        mlflow.artifacts.download_artifacts(model_uri, dst_path=local_dir)
        return local_dir
```

Then you use it like so:
```python
linreg_v1 = Artifact(MlflowRegistryLoader('abc', 'linreg', 'v1'))
linreg_v2 = Artifact(MlflowRegistryLoader('abc', 'linreg', 'v2'))
```

## Suggestions for `Loader` and `Artifact` improvement
But after typing that out, I do agree that it is probably more elegant and a nicer DevX when you just supply the loader to an artifact and hand through the necessary `kwargs` like so:
```python
linreg_v1 = Artifact(MlflowRegistryloader, 'abc', 'linreg', 'v1')
```

Hurdles to adding it to the `loader.load` API are just that the `__init__` of `Artifact` calls `loader.load()` and that is not a method the user is going to override. So we need to be able to handle every possible change in the interface. 

## Passing `loader` to `runner.run`
I think that could be a convenient API. I don't know how prevalent it would be to benchmark multiple models from different sources. Like a local one vs one from a registry. If the local one is far worse, i might not even want to push it to the registry. 

In that case however, I would need two loaders.

The main rationale of separating the artifact and loader logic was that we do have them detangled so you only need to write an artifact class for every artifact type and a loader for every storage location instead of the product of the two. 

I feel a default loader would entangle the two again.

src/nnbench/types.py
maxmynter: We should in any case update the guide/examples when we make another round of changes to the artifact API.

nicholasjng:
Superseded by #120.

PR #111, opened 2024-03-13
Inline the `_cleanup` function for the `FilePathArtifactLoader`. 

maxmynter:
Inline the `_cleanup` function for the `FilePathArtifactLoader`. 

The arguments of `weakref.finalize(obj, func, /, *args, **kwargs)` [should not](https://docs.python.org/3/library/weakref.html#weakref.finalize) own any references to `obj`. Otherwise, the object is never garbage collected.

### DIFF
--- a/src/nnbench/types.py
+++ b/src/nnbench/types.py
@@ -167,12 +167,15 @@ class FilePathArtifactLoader(ArtifactLoader):
     ) -> None:
         self.source_path = str(path)
         if destination:
-            self.target_path = str(Path(destination).resolve())
+            target_path = str(Path(destination).resolve())
             delete = False
         else:
-            self.target_path = str(Path(mkdtemp()).resolve())
+            target_path = str(Path(mkdtemp()).resolve())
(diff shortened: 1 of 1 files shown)


nicholasjng:
Interesting. Why is it used in the standard lib in `tempfile`, then? https://github.com/python/cpython/blob/8e45f7d2aecd8820fc2f144f0bb0159351e46ab1/Lib/tempfile.py#L885-L888

Could you maybe try this in a file? Add a print / log statement to `self._cleanup()` before and after destroying the output of mkdtemp(), both on ordinary exit and with an exception.

Basically, this repro:

```python
# _example.py
from nnbench.types import FilePathArtifactLoader


try:
    f = FilePathArtifactLoader()
    raise RuntimeError("boom")
finally:
    pass
```

and then check for the prints on both branches.

nicholasjng:
Ah, I get it - `_cleanup` and `_rmtree` are static, and thus not bound to the instance they are reaping.

PR #112, opened 2024-03-13
Unique keys

### LINKED ISSUE #99: Improve ensure unique keys in `Contexts` upon merge.
Follow up from #92

- [x] Define and add a behaviour for non-unique keys in `Context.add` and `Context.update`
- [x] Remove the checks for duplicate keys from the `runner.run` method.

maxmynter:
Closes #99 

The `.add` and `.update` methods of the `Context` now check for
key collisions. With the `replace=False` `kwarg` the user can cus-
tomize the behaviour. Upon collision a ValueError is raised.

Remove collision logic from `runner.run` method as it is now hand-
led in the `Context.add` method.

The `.add` method is implemented in term of `.update` to stay DRY.

Change some tests as they contained key collisions.
Updated tests test the desired behaviour based upon the `replace`
flag.

### DIFF
--- a/src/nnbench/context.py
+++ b/src/nnbench/context.py
@@ -88,7 +88,10 @@ class GitEnvironmentInfo:
                 git = "git"
 
             return subprocess.run(  # nosec: B603
-                [git, *args], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding="utf-8"
+                [git, *args],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                encoding="utf-8",
             )
 
         result: dict[str, Any] = {
@@ -246,7 +249,7 @@ class Context:
         str
             Iterator over the context dictionary keys.
         """
-        for k, v in self._ctx_items(d=self._data, prefix="", sep=sep):
+        for k, _ in self._ctx_items(d=self._data, prefix="", sep=sep):
             yield k
 
     def values(self) -> Iterator[Any]:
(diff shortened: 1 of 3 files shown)


src/nnbench/context.py
```diff
@@ -277,26 +280,39 @@ def items(self, sep: str = ".") -> Iterator[tuple[str, Any]]:
...
         ----------
         other : Context
             The other context to update this context with.
+        replace : bool
+            Whether to replace existing context values upon key collision. Raises ValueError otherwise.
+
+        Raises
+        ------
+        ValueError
+            Upon context key collision.
```
nicholasjng, line 310:
```suggestion
            If ``other`` contains top-level keys already present in the context and ``replace=False``.
```

tests/test_context.py
```diff
@@ -71,17 +73,43 @@ def test_context_items():
...
+
+
+def test_add_with_provider_with_key_collision():
+    ctx = Context({"a": 1, "b": 2})
+    with pytest.raises(ValueError, match=".*multiple values for context.*"):
+        ctx.add(lambda: {"a": 3, "c": 4})
+
+
+def test_add_with_provider_with_replace():
+    ctx = Context({"a": 1, "b": 2})
```
nicholasjng, line 103: If `add()` just calls `update()` internally, I'm fine with leaving these two out.

tests/test_context.py
```diff
@@ -71,17 +73,43 @@ def test_context_items():
     assert set(ctx.items()) == expected_items
 
 
+def test_update_with_context_key_collition():
```
nicholasjng, line 76:
```suggestion
def test_context_update_with_key_collision():
```

Same with the other guy (`def test_context_update_duplicate_with_replace`)

nicholasjng (commented):
Nice work!

PR #113, opened 2024-03-13
Add transform proposal

nicholasjng:
Based on the changes contained in #110 (although mostly if not entirely orthogonal).

The proposed `Transform` class is a thin, possibly stateful class mapping
records to other records.

It is generally assumed to be invertible, which is expressed by the
addition of an `iapply()` method (name pending), which can be used to
(approximately) revert records to their old form IF the given transform
supports it.

The added script showcases both proposals, with a barebones example of
why it is useful: To serialize a given model into a set of digestible,
unique identifiers, and doing the reverse in the inverse method.

This way, even custom model classes can be written to JSON, where they
are (very simply) saved via their chosen transform parametrization.

### DIFF
--- a/docs/guides/transforms.md
+++ b/docs/guides/transforms.md
@@ -1,0 +1,53 @@
+# Using transforms to manipulate benchmark records
+
+After a successful benchmark run execution, you end up with your metrics, context, and parameters in a single benchmark record struct.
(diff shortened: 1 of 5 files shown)


nicholasjng:
Question for reviewer(s):

- Should the transform methods become abstract?
- How would you design APIs verifying the transform capabilities hardcoded as bool attributes?

PR #116, opened 2024-03-14
`functools.partial`ize the `nnbench.parametrize` decorators.

### LINKED ISSUE #96: Make `nnbench.parametrize` work with `functools.partial`
You may encounter a situation where you want to parametrize many benchmarks in the same way. 

An example could be label specific metrics like in the `artifact-benchmarking` example #92.
```python
@nnbench.parametrize(
    (
        {"class_label": "O"},
        {"class_label": "B-PER"},
        {"class_label": "I-PER"},
        {"class_label": "B-ORG"},
        {"class_label": "I-ORG"},
        {"class_label": "B-LOC"},
        {"class_label": "I-LOC"},
        {"class_label": "B-MISC"},
        {"class_label": "I-MISC"},
    ),
    tags=("metric", "per-class"),
)
def precision_per_class(
    class_label: str,
    model: Module,
    test_dataloader: DataLoader,
    index_label_mapping: dict[int, str],
    padding_id: int = -100,
) -> float:
...
```

To remove code redundancy, when this exact pattern applies to multiple benchmarks, it would be desirable to use e.g. `functools.partial` on the `nnbench.parametrize` decorator, like so:

```python
per_class_parametrize = functools.partial(nnbench.parametrize, (
        {"class_label": "O"},
        {"class_label": "B-PER"},
        {"class_label": "I-PER"},
        {"class_label": "B-ORG"},
        {"class_label": "I-ORG"},
        {"class_label": "B-LOC"},
        {"class_label": "I-LOC"},
        {"class_label": "B-MISC"},
        {"class_label": "I-MISC"},
    ),
    tags=("metric", "per-class"))

@per_class_parametrize
def precision_per_class(
    class_label: str,
    model: Module,
    test_dataloader: DataLoader,
    index_label_mapping: dict[int, str],
    padding_id: int = -100,
) -> float:
...
```

While the above approach does not throw an error it also does not work. The benchmarks that are parametrized with `functools.partial` are not executed. 

We should investigate that to allow the reuse of code without repetition.

maxmynter:
To use `functools.partial` for `parametrize` (and `product`) it is necessary to call the function returned by `partial`. Other- wise collection does not work.

Add these insights in the guide and example.

Closes #96

### DIFF
--- a/examples/artifact_benchmarking/src/benchmark.py
+++ b/examples/artifact_benchmarking/src/benchmark.py
@@ -1,7 +1,7 @@
 import os
 import tempfile
 import time
-from functools import lru_cache
+from functools import lru_cache, partial
 
 import torch
 from torch.nn import Module
@@ -48,7 +48,8 @@ def precision(model: Module, test_dataloader: DataLoader, padding_id: int = -100
     return torch.mean(precision).item()
 
 
-@nnbench.parametrize(
+parametrize_label = partial(
+    nnbench.parametrize,
     (
         {"class_label": "O"},
         {"class_label": "B-PER"},
@@ -61,7 +62,10 @@ def precision(model: Module, test_dataloader: DataLoader, padding_id: int = -100
         {"class_label": "I-MISC"},
     ),
     tags=("metric", "per-class"),
-)
+)()
+
+
+@parametrize_label
 def precision_per_class(
     class_label: str,
     model: Module,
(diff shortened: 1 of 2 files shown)


examples/artifact_benchmarking/src/benchmark.py
```diff
@@ -60,8 +59,10 @@ def precision(model: Module, test_dataloader: DataLoader, padding_id: int = -100
         {"class_label": "B-MISC"},
         {"class_label": "I-MISC"},
     ),
-    tags=("metric", "per-class"),
-)
+    tags=("metric", "per-class"))()
```
maxmynter, line 62: The last call `()` was missing in previous implementations and was what broke the collection in the version previously.

examples/artifact_benchmarking/src/benchmark.py
```diff
@@ -61,7 +62,10 @@ def precision(model: Module, test_dataloader: DataLoader, padding_id: int = -100
         {"class_label": "I-MISC"},
     ),
     tags=("metric", "per-class"),
-)
+)()
```
maxmynter, line 65: This last call `()` was missing in previous implementations and was what broke the collection in the version previously.

nicholasjng: Ah yes, because we removed the no-parentheses fastpath for `parametrize` and product.

nicholasjng:
Call it `parametrize_label` and you're good!

PR #119, opened 2024-03-19
Check loaded module names before importing a new one from a file

### LINKED ISSUE #115: Module path construction causes typecheck fails
Raised internally. Repro:

```python
import nnbench
from nnbench import types, context
import pandas as pd
from pathlib import Path

class MyDataSet(types.Artifact):
	def deserialize(self) -> None:
		path = Path(self.path).resolve()
		self._value = pd.read_csv(path, sep="\t")
		

@nnbench.benchmark()
def my_benchmark(dataset: MyDataSet) -> float:
	return 1.0

if __name__ == "__main__":
	dataset = MyDataSet(types.LocalArtifactLoader(path="ames_housing.csv"))
	
	runner = nnbench.BenchmarkRunner()
	reporter = nnbench.BenchmarkReporter()

	results = runner.run("example.py", params={
	                     	"dataset": dataset,
	                     })
	reporter.display(results) 
```

Error:

```
(.venv) janwillem@macbook-janwillem [nnbench-trial2] % python example.py
Traceback (most recent call last):
  File "/Users/janwillem/Developer/hack/nnbench-trial2/example.py", line 27, in <module>
    results = runner.run("example.py", params={
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/janwillem/Developer/hack/nnbench-trial2/.venv/lib/python3.11/site-packages/nnbench/runner.py", line 239, in run
    self._check(dparams)
  File "/Users/janwillem/Developer/hack/nnbench-trial2/.venv/lib/python3.11/site-packages/nnbench/runner.py", line 130, in _check
    raise TypeError(
TypeError: expected type <class 'example.py.MyDataSet'> for parameter 'dataset', got <class '__main__.MyDataSet'>
```

This means that the module loading will screw up some of the types. (If I understand correctly, it expects a `MyDataSet` from the target file, but gets one from the currently executed module, i.e. `__main__`).

The module import name is obviously a bug, but the general behavior might be expected, although surprising. Possible actions here include making a warning/note on using `__main__` as `path_or_module` qualifier when running in-file benchmarks.

nicholasjng:
Fixes a bug that prevents in-file benchmark execution due to a type mismatch when calling runner.run() with __file__ as argument (or just the name).

The reason was that on encountering a file name, the Python module that the file pointed to was imported under its absolute path name regardless of whether it was already present in sys.modules (in this case, it was, because it is also __main__).

Adds a check in `import_file_as_module` whether a module already exists in sys.modules for the given file name, in which case that module is returned.

Fixes #115.

### DIFF
--- a/src/nnbench/util.py
+++ b/src/nnbench/util.py
@@ -30,21 +30,27 @@ def ismodule(name: str | os.PathLike[str]) -> bool:
 
 def modulename(file: str | os.PathLike[str]) -> str:
     """Convert a file name to its corresponding Python module name."""
-    fpath = Path(file)
+    fpath = Path(file).with_suffix("")
     if len(fpath.parts) == 1:
         return str(fpath)
 
-    filename = fpath.with_suffix("").as_posix()
+    filename = fpath.as_posix()
     return filename.replace("/", ".")
 
 
 def import_file_as_module(file: str | os.PathLike[str]) -> ModuleType:
-    fpath = Path(file)
+    fpath = Path(file).resolve()  # Python module __file__ paths are absolute.
     if not fpath.is_file() or fpath.suffix != ".py":
         raise ValueError(f"path {str(file)!r} is not a Python file")
 
-    # TODO: For absolute paths, the resulting module name will be horrifying
-    #  -> find a sensible cutoff point (project root)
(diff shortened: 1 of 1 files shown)


nicholasjng:
fyi, this makes the repro from #115 execute just fine.

Issue #126, opened 2024-03-21
Add global memo cache and eviction API


nicholasjng:
As per this TODO: https://github.com/aai-institute/nnbench/blob/65fc45b3ef3f3992062aea796e72ac4dc1d66fcd/src/nnbench/types/types.py#L104-L119

Steps are

1 - bring back the global cache + lock,
2 - insert values based on the id of the inserting memo,
3 - add a `clear_memo_cache` API, maybe factoring out everything related into a `nnbench.types.memo` submodule.
4 - (optional) - check if evicting the memoized value in Memo.__del__() does anything (debug logs!)
5 - drop functools.cache decorators everywhere (examples!)

NB: Check that the `Memo.__call__()` type annotation is not affected by the solution, since that breaks typechecking in benchmarks. (With a decorator, this should not be a problem.)

What would be super nice: Showcase a parametrized benchmark with a collection of large numpy array memos, checking if evicting an array after the last run of the family and triggering GC (`gc.collect()`) does anything. memray is (hopefully) your friend: https://bloomberg.github.io/memray/index.html
-> This needs #125.

PR #127, opened 2024-03-22
State injection into setup and teardown

### LINKED ISSUE #125: Add `nnbench.State` object holding current benchmark information, inject into setup/teardown tasks
We want to experiment with a global memoization cache, which we want to explicitly clear in a teardown task after a family of benchmarks for a model (say NER on distilbert) has run.

This means that the setup and teardown tasks need to know which benchmark they are currently applied in.

For a single benchmark, say

```python
@nnbench.benchmark
def echo(s: str) -> str:
    print(s)
```

the corresponding `State` should look something like this:

```python
def tearDown(state, **params): # <- or maybe a mappingproxy of the params?
    print(state)

# name: "echo"
# family: "echo" <- this is for parametrized benchmarks, where the family name equals the function name.
# family_size: 1 <- how many members in the current benchmark family?
# family_index: 0 <- which number of the family is it?
```

More metadata suggestions welcome. After this, we can try evicting a memo from the cache on the condition `family_index == family_size - 1`.

maxmynter:
Closes #125 

Trailing ToDos (actual work with the `State` Object in setup and teardown tasks will be addressed in the PR for #126  .

### DIFF
--- a/src/nnbench/core.py
+++ b/src/nnbench/core.py
@@ -10,6 +10,7 @@ import warnings
 from typing import Any, Callable, Iterable, Union, get_args, get_origin, overload
 
 from nnbench.types import Benchmark
+from nnbench.types.types import NoOp
 from nnbench.types.util import is_memo, is_memo_type
 
 
@@ -50,10 +51,6 @@ def _check_against_interface(params: dict[str, Any], fun: Callable) -> None:
 
 def _default_namegen(fn: Callable, **kwargs: Any) -> str:
     return fn.__name__ + "_" + "_".join(f"{k}={v}" for k, v in kwargs.items())
-
-
-def NoOp(**kwargs: Any) -> None:
-    pass
 
 
 # Overloads for the ``benchmark`` decorator.
@@ -178,7 +175,14 @@ def parametrize(
                 )
             names.add(name)
 
-            bm = Benchmark(fn, name=name, params=params, setUp=setUp, tearDown=tearDown, tags=tags)
+            bm = Benchmark(
+                fn,
+                name=name,
+                params=params,
+                setUp=setUp,
+                tearDown=tearDown,
+                tags=tags,
+            )
             benchmarks.append(bm)
         return benchmarks
 
@@ -236,7 +240,14 @@ def product(
                 )
             names.add(name)
 
-            bm = Benchmark(fn, name=name, params=params, setUp=setUp, tearDown=tearDown, tags=tags)
+            bm = Benchmark(
+                fn,
+                name=name,
+                params=params,
+                setUp=setUp,
+                tearDown=tearDown,
+                tags=tags,
+            )
             benchmarks.append(bm)
         return benchmarks
 
--- a/src/nnbench/runner.py
+++ b/src/nnbench/runner.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import collections
 import contextlib
 import inspect
 import logging
@@ -15,7 +16,7 @@ from pathlib import Path
 from typing import Any, Callable, Generator, Sequence, get_origin
 
 from nnbench.context import Context, ContextProvider
-from nnbench.types import Benchmark, BenchmarkRecord, Parameters
+from nnbench.types import Benchmark, BenchmarkRecord, Parameters, State
 from nnbench.types.util import is_memo, is_memo_type
 from nnbench.util import import_file_as_module, ismodule
 
@@ -247,6 +248,9 @@ class BenchmarkRunner:
         if not self.benchmarks:
             self.collect(path_or_module, tags)
 
+        family_sizes: dict[str, Any] = collections.defaultdict(int)
+        family_indices: dict[str, Any] = collections.defaultdict(int)
+
         if isinstance(context, Context):
             ctx = context
         else:
@@ -258,6 +262,9 @@ class BenchmarkRunner:
         if not self.benchmarks:
             warnings.warn(f"No benchmarks found in path/module {str(path_or_module)!r}.")
             return BenchmarkRecord(context=ctx, benchmarks=[])
+
+        for bm in self.benchmarks:
+            family_sizes[bm.fn.__name__] += 1
 
         if isinstance(params, Parameters):
             dparams = asdict(params)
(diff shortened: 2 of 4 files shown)


src/nnbench/core.py
```diff
@@ -224,7 +239,8 @@ def decorator(fn: Callable) -> list[Benchmark]:
         benchmarks = []
         names = set()
         varnames = iterables.keys()
-        for values in itertools.product(*iterables.values()):
+        cartesian_product = list(itertools.product(*iterables.values()))
```
maxmynter, line 242: `list` here forces execution of the generator. I think that is no problem as the number of benchmarks will not be too computationally extensive. Additionally, we nee the number of values in the cartesian product as well as the index anyway, so we might as well unravel the generator here instead of in the following for loop.

src/nnbench/types/types.py
```diff
@@ -164,11 +175,41 @@ class Benchmark:
...
+            super().__setattr__(
+                "state",
+                State(
+                    name=self.name or "",
+                    function=self.fn,
+                    family=self.fn.__name__,
+                    family_size=1,
+                    family_index=0,
+                ),
+            )
```
maxmynter, line 194: Is it sufficiently clear that this is the default if the Benchmark state is not set? that is the case for standalone benchmarks (which do not come out of `parametrize` or `product`). Therefore `family_size=1` and `family_index=0`.

nicholasjng (commented):
I think the state should not be tied to the benchmark at all. Right now, the dependency is `Benchmark -> State` (as in, the state is a member of the benchmark struct), and State even has a reference to the benchmark function.

What I am looking for is more an on-the-fly constructed POD struct (i.e. holding data only, to be discussed) in `runner.run()`, mapping a benchmark to its corresponding state irrespective of its origin.

You can infer name, family name, and index from the benchmark data no problem, and you just need to bookkeep a small map of family name -> index.

But as for the family size, you're right that this will be needed before running the benchmarks. The single benchmark case is easy (size 1), with the parametrized ones it's a little harder.

Maybe the `@parametrize/@product` decorators should return a `BenchmarkFamily` object, which holds its size as a property?

But for now, just go through the `self.benchmarks` list and populate the family sizes like so:

```python
family_sizes = collections.defaultdict(int)
family_idxs = collections.defaultdict(int)

for bm in self.benchmarks:
    family_sizes[bm.fn.__name__] += 1
    ...

# later, in execution loop:
state = State(..., family_size=family_sizes[bm.fn.__name__], family_idx=family_idxs[bm.fn.__name__])
# advance index after state dispatch.
family_idxs[bm.fn.__name__] += 1
```

src/nnbench/core.py
```diff
@@ -178,7 +178,14 @@ def decorator(fn: Callable) -> list[Benchmark]:
                 )
             names.add(name)
 
-            bm = Benchmark(fn, name=name, params=params, setUp=setUp, tearDown=tearDown, tags=tags)
+            bm = Benchmark(
```
nicholasjng, line 181: Did ruff change this formatting?

src/nnbench/runner.py
```diff
@@ -16,6 +17,7 @@
 
 from nnbench.context import Context, ContextProvider
 from nnbench.types import Benchmark, BenchmarkRecord, Parameters
+from nnbench.types.types import State
```
nicholasjng, line 20: Please import `State` into `nnbench.types`, and add the import to L19 above.

src/nnbench/runner.py
```diff
@@ -247,6 +249,12 @@ def run(
         if not self.benchmarks:
             self.collect(path_or_module, tags)
 
+        family_sizes: dict[str, Any] = collections.defaultdict(int)
+        for bm in self.benchmarks:
+            family_sizes[bm.fn.__name__] += 1
+
+        family_indices: dict[str, Any] = collections.defaultdict(int)
+
```
nicholasjng, line 257: I would put this below the `if not self.benchmarks` check in L266, because we don't need to run it if there are no benchmarks.

src/nnbench/runner.py
```diff
@@ -274,6 +282,14 @@ def _maybe_dememo(v, expected_type):
             return v
 
         for benchmark in self.benchmarks:
+            bm_family = benchmark.fn.__name__
+            bm_state = State(
+                name=benchmark.name or bm_family,
```
nicholasjng, line 287: `benchmark.name` should never be None (or ""), so you can leave out the or branch.

src/nnbench/runner.py
```diff
@@ -291,14 +307,14 @@ def _maybe_dememo(v, expected_type):
                 "parameters": bmparams,
             }
             try:
-                benchmark.setUp(**bmparams)
+                benchmark.setUp(bm_state, **bmparams)
```
nicholasjng, line 310: nit: I would name it `state`, looks nicer in the code flow.

src/nnbench/types/types.py
```diff
@@ -170,6 +172,24 @@ def __post_init__(self):
...
+
+        original_tearDown = self.tearDown
+
+        def wrapped_tearDown(state: State, /, *args: Any, **kwargs: Any) -> None:
+            # TODO: tearDown logic
+            print("tearDown: ", state)
+            original_tearDown(*args, **kwargs)
+
+        super().__setattr__("tearDown", wrapped_tearDown)
+
```
nicholasjng, line 192: Ah, no - I meant we should change the setUp/tearDown calling convention, as in, require callables of signature `(state, params) => Any`, not wrap the callable itself.

(We will need to adjust the docs on setUp and tearDown tasks afterwards.)

NB: The `NoOp` from that module must then become `def NoOp(state, params): pass`.

I guess it is nicer to pass the dict instead of the kwargs. But this requires wrapping it into a form where the user cannot inject any parameters, e.g. `types.MappingProxy`.

nicholasjng (commented):
Nice! We're almost there.

src/nnbench/core.py
maxmynter: Ah, yes. The linebreak was introduced when i added the state. Apparently ruff did not remove the linebreak even after I removed the passing of `State`.

src/nnbench/runner.py
maxmynter: I had mypy complaining here. But I could get rid of that warning by changeing `if not self.name ` to `if self.name is None` in the benchmark `__post_init__` where we create the name if not set.

src/nnbench/types/types.py
```diff
@@ -6,24 +6,23 @@
...
 
-def NoOp(**kwargs: Any) -> None:
+def NoOp(state: State, params: Mapping[str, Any] = MappingProxyType({})) -> None:
     pass
 
 
+class CallbackProtocol(Protocol):
+    def __call__(self, state: State, params: Mapping[str, Any]) -> None: ...
+
+
```
nicholasjng, line 25: I would just go with `Callable[[State, Mapping[str, Any]], None]`, unless I'm missing something?

nicholasjng (commented):
Nice work!

PR #129, opened 2024-03-25
Add new memoization guide

nicholasjng:
Follows up the artifact guide with an example of a memo. This is work in progress until the cache and eviction policy are implemented and confirmed to work.


maxmynter (approved):
Nice! 
This PR also clears some questions I had regarding the implementation of the Memo. I'll have to adapt #130 a but but will get it in a reviewable state today.

PR #130, opened 2024-03-25
Add global memo cache and integrate with the setUp and teardown injection

### LINKED ISSUE #126: Add global memo cache and eviction API

maxmynter:
Closes #126 
Needs #127
- [ ] ~~Integrate Cache into setup and teardown based on state~~ (This is to be done by the user in the setUp and tearDown hooks).


nicholasjng:
Before you continue: Let's try with a bare `dict` with lock in `nnbench.types.memo` as a cache first, since all methods you implemented are class-scoped.

Likewise, getting a memo from cache should prompt a key lookup first:

```
# this in a decorator applied to Memo.__call__():
_sid = id(self)
if _sid in _memo_cache:
    return _memo_cache[_sid]
return self.__call__()
```

Issue #145, opened 2024-07-18
macOS test failure due to GitHub Apple Silicon oddities


maxmynter:
We are getting a failing test

```bash
FAILED tests/test_context.py::test_cpu_info_provider - RuntimeError: 'pmgr' entry was not found in AppleARMIODevice service
```

The root cause for this seems to be the GitHub VM that runs the testing environment as per [this discussion](https://github.com/giampaolo/psutil/issues/2382).

Locally tests run smoothly.

PR #146, opened 2024-07-18
Safe setting of CPU Info

maxmynter:
as the context information might not be available on all devices, e.g. in our GitHub CI VM the CPU frequency.

Wrap setting of dict entry in a try, except blog that throws warnings when a value cannot be obtained.

This is to address the CI failure described in #145 and very much a suggestion that we can discuss. 

In any case, I think a failure of getting context info should not break the test execution in case they take some time.
As a data scientist using nnbench, I would want our suite to run through so i can react to failures. However, I definitely want information about missing values such that i can implement fixes to these in case they are crucial.

### DIFF
--- a/src/nnbench/context.py
+++ b/src/nnbench/context.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import itertools
+import logging
 import platform
 import sys
 from typing import Any, Callable, Iterator, Literal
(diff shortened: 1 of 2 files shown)


nicholasjng (commented):
I'd be in favor of a more tailored solution for the problem rather than having a `None` fallback for everything - `cpu_freq()` only seems to fail on macOS, and then we could always set frequency/min/max all to 0.

For example, if reading `platform` values really fails, you've got much bigger problems than running nnbench. So I'd prefer a try-except just around cpu_freq() if possible.

nicholasjng (approved):

PR #149, opened 2024-10-10
Remove requirement files, strict version pins from examples

nicholasjng:
In preparation of running a cronjob in GitHub Actions that checks verifies the examples to work.

Side effect: This stops the dependabot alert spam.

I would like to add a job to our CI that runs all examples on a schedule, e.g. every day at midnight.

Updates the Python pins to 3.10+, since nnbench is set up that way now.


nicholasjng:
What we could do is add lockfiles for a quick install+run workflow with uv, otherwise we just tell people to `[uv] pip install .` like I think we're doing already at this point.

PR #152, opened 2024-10-24
Add CLI entrypoints for module and program invocation

nicholasjng:
Supports `nnbench <benchmark> <options>` as well as `python -m nnbench <benchmark> <options>`.

Like pytest, we give the option to omit the target location, in which case it becomes a directory named "benchmarks".

Supports a very rudimentary context building from switches method, and tag filtering via switch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,9 @@ docs = [
     "neoteroi-mkdocs",
 ]
 
+[project.scripts]
+nnbench = "nnbench.cli:main"
+
 [tool.setuptools]
 package-dir = { "" = "src" }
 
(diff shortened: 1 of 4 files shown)


nicholasjng:
Test setup:

```python
# example.py

import nnbench

# no idea yet how to hand off args on the command line, so we need them hardcoded.
@nnbench.benchmark
def add(a: int = 1, b: int = 2) -> float:
    return a + b
```

Then, `nnbench example.py` or `python -m nnbench example.py` does the trick. Bonus: If you put the file into a directory named "benchmarks", even `nnbench` without any args works.

PR #155, opened 2024-10-26
Add Python 3.13 tests to CI

nicholasjng:
Since it's been released, we should start testing it as well.

---------

Same problem as in lakefs-spec, we rely on pyarrow, which does not have 3.13 wheels yet.


nicholasjng:
Huh, numpy dropped out of the lockfile. Did pyarrow drop the dependency on it?

AdrianoKF:
> Huh, numpy dropped out of the lockfile. Did pyarrow drop the dependency on it?
> 
> EDIT: yes it did.

https://github.com/apache/arrow/issues/25118 - seems so!

PR #156, opened 2024-10-30
Remove setuptools_scm in favor of explicit version management

nicholasjng:
setuptools_scm has a few drawbacks to version management by hand, such as inflexibility for releasing off-tag, and missing info on shallow checkouts.

As such, revert back to a static version number, with tagged commits as version bumps.


nicholasjng:
To get a dev version of nnbench to appear in the lockfile, we could tag a development version with a `dev` suffix (maybe not numbered). Then we can access the actual revision as `git rev-parse HEAD`.

PR #164, opened 2024-11-18
Add `nnbench compare` subcommand, comparison function

nicholasjng:
Works using metric (benchmark) names as columns, and records as lines. Alternatively, we could support metrics as rows and record IDs as columns in the future.

Contains logic to add parameter values and context values to the table as columns, also directly on the command line. In the future, we could make a comparison class that can be subclassed to avoid option clutter on the command line for customizing the rich table display.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,11 @@ classifiers = [
     "Typing :: Typed",
 ]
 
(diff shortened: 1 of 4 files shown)


nicholasjng:
Merging this now, the (numerous) follow-ups can be tackled in other issues, but don't have to be implemented for the next release yet.

PR #165, opened 2024-11-19
Refactor console reporter

nicholasjng:
This also changes the way that context values are reported, they now
appear as key-value pairs above the table.

Each benchmark metric is one row, with the value appearing underneath.
A column-first mode could be added later.

Changes the inheritance structure to not put a `display()` method on the
base class.

Drops tabulate from requirements in favor of rich.


nicholasjng:
Example:

```console
.venv ➜ nnbench run .sandbox/example.py --context="foo=bar"
foo: bar
┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Benchmark ┃ Value                       ┃ Wall time (ns) ┃ Parameters       ┃
┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ add       │ 3                           │ 1667           │ {'a': 1, 'b': 2} │
│ mul       │ 2                           │ 500            │ {'a': 1, 'b': 2} │
│ sub       │ ERROR: 'sub' is unsupported │ 5625           │ {'a': 1, 'b': 2} │
└───────────┴─────────────────────────────┴────────────────┴──────────────────┘

```

PR #168, opened 2024-11-20
Some tweaks related to context ingestion

nicholasjng:
Unfortunately, the first prototype for builtin context provider support from the command line was not successful.

The alternative is to provide values as
key=`from nnbench.context import xy; print(xy()())`, but that returns the serialized JSON struct, which is not all that useful. Maybe we can get around that by optionally processing JSON-like values into the context.


nicholasjng:
This is stale - most of the content here has found its way into `main` in the latest TOML parsing addition.

PR #179, opened 2024-12-02
Add nnbench CLI guide

nicholasjng:
Contains all commands, their usages and use scenarios, and an appendix section on how to configure the CLI experience via `pyproject.toml`.

------------------

I also made two small tweaks to the CLI, namely printing help when calling `nnbench` without any arguments, and displaying the benchmark run name in comparison tables rendered by `nnbench compare`.


docs/guides/cli.md
```diff
@@ -0,0 +1,293 @@
...
+  -h, --help            show this help message and exit
+  -P, --include-parameter <name>
+                        Names of input parameters to display in the comparison table.
+  -C, --include-context <name>
+                        Context values to display in the comparison table. Use dotted syntax for nested context values.
+  -E, --extra-column <name>
+                        Additional record data to display in the comparison table.
+```
+
+Supposing we have the following records from previous runs:
```
maxmynter, line 81: We could shorten this example to use only one variable, e.g. `add`. I feel using three makes this example pretty long without adding much additional information

docs/guides/cli.md
```diff
@@ -0,0 +1,293 @@
...
+$ nnbench compare record1.json record2.json
+┏━━━━━━━━━━━━━━━━━━┳━━━━━┳━━━━━━┳━━━━━┓
+┃ Benchmark run    ┃ add ┃ mul  ┃ sub ┃
+┡━━━━━━━━━━━━━━━━━━╇━━━━━╇━━━━━━╇━━━━━┩
+│ nnbench-f0822b3c │ 75  │ 1250 │ 25  │
+│ nnbench-4f804c24 │ 75  │ 1250 │ 25  │
+└──────────────────┴─────┴──────┴─────┘
+```
+
+To include benchmark parameters in the table, use the `-P` switch (you can supply this multiple times to include multiple parameters):
```
maxmynter, line 205: This refers to the arguments we pass in the benchmarks, right? 

If so, we should half a sentence explaining that as I had to think a bit about it. Someone new to nnbench may not make that connection at all.

docs/guides/cli.md
```diff
@@ -0,0 +1,293 @@
...
+                        Additional context values giving information about the benchmark run.
+  -t, --tag <tag>       Only run benchmarks marked with one or more given tag(s).
+  -o, --output-file <file>
+                        File or stream to write results to, defaults to stdout.
+```
+
+So to run a benchmark workload contained in a single `benchmarks.py` file, you would run `nnbench run benchmarks.py`.
+For tips on how to structure and annotate your benchmarks, refer to the [organization](organization.md) guide.
+
+For injecting context values on the command line, you need to give the key-value pair explicitly by passing the `--context` switch.
```
maxmynter, line 50: I think it makes sense to also refer to the custom implementation that comes later. Otherwise me might loose someone looking for this information and - after having found this here - not reading further in the guide.

nicholasjng:
Agree 100% on all points - will add them over the course of the evening.

PR #191, opened 2024-12-18
Run pre-commit autoupdate, rename decorator test file

nicholasjng:
As stated in the title.


nicholasjng:
Since memray is not supported on Windows, we have to mark all upcoming memory usage tests with `pytest.mark.skipif(platform.system == "Windows", ...)`. 

Since memray is not even installable on Windows, it might require moving everything to a proper test module and skipping the complete module if we're on Windows, to avoid import errors.

PR #208, opened 2025-02-06
fix: Use walk() to traverse more than one directory in benchmark collection

nicholasjng:
Previously, when inputting a directory for benchmarking, only benchmarks at that directory level were collected because of the use of `Path.iterdir()`.

However, if we want to support arbitrary depth in a benchmark directory layout, we need to use depth traversal, which is facilitated either by os.walk() or, for 3.12 and newer, Path.walk().

Changes both callsites (CLI+multiprocessing and `nnbench.collect()`) to the new `all_python_files()` API.


nicholasjng:
Test failure is unrelated.

PR #212, opened 2025-02-12
chore: Upgrade lockfile dependencies, pre-commit hooks

nicholasjng:
I can't reproduce the memray failure on Python 3.11 locally, maybe this has to do with a stale NumPy?


nicholasjng:
Huh, interesting. Wonder what the reason was there.

PR #213, opened 2025-02-12
Add MLFlowIO to enable MLflow logging of benchmark results

nicholasjng:
Introduces a `BenchmarkServiceIO` protocol, establishing an interface for streaming JSON results to web services like MLflow, Grafana, WandB, ...

Also introduces multiple dispatch of IO based on the value of `outfile`, i.e. the value of the `-o` switch.

The introduced taxonomy is stdout | file | database | service, and loads a different IO class depending on the detected type of IO requested.

For files, inferring the IO type is a bit harder, since local files are usually given without a leading protocol. For this case, we look at the file extension as well.

Every other IO is detected by the leading protocol of the given URI, so `mlflow://experiment/run/subrun` leads to the MLflow IO being requested.

-------------------------

Usage (create two files, `example.py` and `conf.py`, with the following contents limited by horizontal bars):

```
# -----------------------
# file: example.py
# -----------------------

import nnbench


@nnbench.benchmark
def add(a: int, b: int) -> int:
    return a + b


@nnbench.benchmark(name="My super summation")
def mul(a: int, b: int) -> int:
    return a * b

# -------------------
# file: conf.py
# -------------------

def a() -> int:
    return 200


def b() -> int:
    return 100
```

Then, `nnbench run example.py --context=foo=bar -o mlflow://test/my-run/my-subrun` streams the results into the given nested subrun, under the "test" experiment (which you may have to create beforehand in the UI). You will also have to set `MLFLOW_TRACKING_URI=http://localhost:$PORT` in the shell you're executing from.

Screenshots:

Screenshot 2025-02-12 at 16 35 48

Screenshot 2025-02-12 at 16 36 11

cc @janwillemkl

### DIFF
--- a/src/nnbench/cli.py
+++ b/src/nnbench/cli.py
@@ -15,7 +15,7 @@ from typing import Any
 from nnbench import ConsoleReporter, __version__, collect, run
 from nnbench.config import NNBenchConfig, parse_nnbench_config
 from nnbench.context import Context, ContextProvider
-from nnbench.reporter import FileReporter
+from nnbench.reporter import FileReporter, IOType, get_io_type
 from nnbench.types import BenchmarkRecord
 from nnbench.util import all_python_files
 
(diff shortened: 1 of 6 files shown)


nicholasjng:
Summarizing a few of my thoughts on this:

- I like the multiple dispatch, but right now it's needlessly complicated in its architecture, because we save file IOs for different file extensions instead of storage systems.
- To unify this, we would either have to make a few shallow wrappers for each supported file system protocol, or make a monolithic class and register it for all protocols (fsspec can already handle all protocols, so we don't need more than one impl)
- If we unambiguously dispatch on protocol, we can drop the multiple registries, and handle registration / deregistration directly in `src/nnbench/reporter/__init__.py`. I plan on handling that next.

Specifically for mlflow, the protocol is so far working well, but there are a few quirks:

1. The code does require a tracking URI to be set, either through an envvar (`MLFLOW_TRACKING_URI`) or through a constructor argument. I would like to address this by giving proper state to the reporter in its `__init__()`, and allowing the CLI user to configure arguments via pyproject.toml.
2. If I don't put a subrun, MLflow still creates more runs under the selected run, autogenerating a name - what's up with that? I'm tempted to just put the nnbench run name as mlflow run name to be able to pipe more state directly into mlflow.
3. I haven't found a way to inject more metadata (descriptions of benchmarks, i.e. metrics), and I still have to check what happens when you're trying to log parameters.

tl,dr: Happy with how it's going, more to come soon to make the experience better and the code easier to grasp.

PR #218, opened 2025-02-18
docs: Improve CLI reference and configuration guide

nicholasjng:
Fixes a few outdated guides, and improves the CLI reference by introducing some metavars and descriptions.


nicholasjng:
NB: I would like to trim the accuracy floats in the table to a predefined number of significant digits, this would most likely work with some TOML config.

